PR #67, opened 2020-08-04
Feature/wbs1.6/weekly summaries

AndonIgnatov:
Please review and merge the "feature/wbs1.6/summary section" branch. I've made sure to sync my branch with the latest "development branch so there won't be any conflicts.


src/components/WeeklySummary/__mocks__/weeklySummaryMockData.js
```diff
@@ -0,0 +1,45 @@
+
+import moment from 'moment';
+import 'moment-timezone';
+
+const dueDateThisWeek = moment().tz('America/Los_Angeles').endOf('week').toISOString();
+const dueDateLastWeek = moment().tz('America/Los_Angeles').endOf('week').subtract(1, 'week')
+  .toISOString();
+const dueDateBeforeLast = moment().tz('America/Los_Angeles').endOf('week').subtract(2, 'week')
+  .toISOString();
```
AGCapdeville, line 9: Here you set the time-zone to 'America/Los_Angeles'. Should this be set as a standard in global?

src/components/WeeklySummary/WeeklySummary.jsx
```diff
@@ -0,0 +1,348 @@
...
+      summary: '',
+      summaryLastWeek: '',
+      summaryBeforeLast: '',
+      mediaUrl: '',
+      weeklySummariesCount: 0,
+      mediaConfirm: false
+    },
+    dueDate: moment().tz('America/Los_Angeles').endOf('week').toISOString(),
+    dueDateLastWeek: moment().tz('America/Los_Angeles').endOf('week').subtract(1, 'week').toISOString(),
+    dueDateBeforeLast: moment().tz('America/Los_Angeles').endOf('week').subtract(2, 'week').toISOString(),
```
AGCapdeville, line 36: .tz('America/Los_Angeles') 
Should these be set? Or should this be referenced from a global const and passed down?

src/components/WeeklySummary/__mocks__/weeklySummaryMockData.js
AndonIgnatov: Thanks Adam, originally it was supposed to be local time to the user but after discussion with Jae he wants it LA time. If we did local, things can get really complicated, we ran a few scenarios at the time of discussion but keeping it LA time would be the simpler approach which works for Jae as well. Keep in mind that regardless of which timezone you start the week, it will end in that same timezone so you're not going to lose or gain any time, it's the same. Just the countdown timer will be showing you the time left relative to LA time.

src/components/WeeklySummary/WeeklySummary.jsx
AndonIgnatov: Good question, yes I need them set as fallback to cover the case when for whatever reason the dueDates are not coming from the database. You can see when I use them in `componentDidMount()`. There are various scenarios that this could happen, so without that bit the component might not work properly, it needs the due dates, so in those rare cases pull local time. Think of it as fallback. In general local time obtained through moment.js is taken from the local machine which makes it vary unreliable, you could set your computer's date to 5 yrs in the past or in the future even though the odds of this happening in real life are low it could happen. So in the component I make sure to always pull those dates from the server **first**, this is absolutely must but if for whatever reason those dates aren't there I pull local time, as simple as that.

PR #71, opened 2020-08-18
test: Unit testing for Dashboard and Login

chrisweilacker:
Fixed Unit Testing on Dashboard and log in.
Implemented several redirection tests for the Login Page including New User, Dashboard, and going to login from another page.


AndonIgnatov:
Hi Chris, noticed your yarn.lock fine is downgrading some packages. Was that intentional?

chrisweilacker:
> Hi Chris, noticed your yarn.lock fine is downgrading some packages. Was that intentional?

In the instructions for getting the app up and running, it had mentioned not ignoring and fixing some of the errors/warnings with regard to eslint by downgrading the package. I believe if I remember correctly eslint-airbnb needed a lower version of eslint as a dependency to work.  There were also warnings thrown for popper/jquery so I went ahead and added those as well.

AndonIgnatov:
OK, cool. I think Nithesh has been assigned to look into the whole ESLint issue as many members were complaining about issues. Regarding popper.js I also ran into a warning when testing reactstrap tooltips. I researched the issue and implemented the solution here: https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/blob/development/src/__tests__/__mocks__/popper.js.js You can find out more about it in the comment in that file. That seams to work. but I'm not sure about the warnings you were getting it might be a different issue.
I'll try to run your tests today and will report back here. BTW, in general for more detailed discussions we use Slack https://highest-good.slack.com/archives/C010UK6NQ84 So, in the future when you have a new PR please make sure to post it also on Slack in that channel so there could be a discussion there if needs be. Thx

AndonIgnatov:
OK, I just ran your tests. All looking good except a couple of small things.
I also checked the coverage and it's looking great.

1. src/components/Dashboard/Dashboard.test.jsx 
✔ This one is a PASS.
2. src/components/Login/Login.test.jsx (1 error)
This is the error I'm getting, it's just on the last test:
image
3. src/__tests__/Logout.test.js ✔ 
✔ This one is a PASS.
4.  src/__tests__/LoginPage.test.js 
This one is a PASS but the test 'should redirect to dashboard if no previous redirection' is causing this warning:
image
I checked that test code and it's pretty straight forward, you're just simulating a successful login with no history.
I'm not sure why that warning is, I tried reproducing this by doing an actual login but I'm not getting that warning in the Console. Interestingly though there is a similar one when you do a logout. I'm not sure what the current state of the login functionality is, I think there might be someone still working on it or will be worked on in near future. I'll have to ask Jae on that.
Are you also getting that warning when you run this test?

chrisweilacker:
Yes, I also have an error pop up. I had made a note of it to ask later as I only got the error when accessing the dashboard, so don't think it has something to do with the tests.  It seems the error occurs on the exit of the weekly summaries section, the only unmount code I see throughout the app is to clearErrors() but don't see it in the Weekly Summaries section.  I pushed up a commit just now to clear the error in the Login.test.jsx, I had accidentally deleted one of the import lines when I was cleaning up some code.

AndonIgnatov:
Yes, the logout warning is gone now, thanks for clearing that :)

AndonIgnatov:
Hey Chris, you might also want to update that snapshot, it's changed since your last push:
image

PR #74, opened 2020-08-27
Unit testing Leaderboard, Header, and ForcePasswordUpdate

chrisweilacker:
Completed Unit Testing for Leaderboard, Header, and FocePasswordUpdate.

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -22,7 +22,7 @@ import {
 } from 'reactstrap'
 import { UserRole } from '../../utils/enums'
 
-class Header extends React.Component {
+export class Header extends React.Component {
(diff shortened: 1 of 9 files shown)


src/components/Header/Header.jsx
```diff
@@ -22,7 +22,7 @@ import {
 } from 'reactstrap'
 import { UserRole } from '../../utils/enums'
 
-class Header extends React.Component {
+export class Header extends React.Component {
```
Junhao6503, line 25: I think we don't need to export the un-connected header component.

chrisweilacker: That is used in the Header.test.jsx file to test some of the page structure.

PR #83, opened 2020-09-02
Feature/wbs0.1

vnseattle:
Solved the problem deleting wbs


chrisweilacker (approved):
Looks good everything seems to be working now.

PR #89, opened 2020-09-11
added test cases for higher test coverage for Timelog and UserManagement

Junhao6503:
The coverage for Timelog is still low due to failing to interact with TinyMCE. Did a lot of research yesterday, but got no luck.


chrisweilacker (approved):
Looks good.

PR #91, opened 2020-09-16
allow adding of why, intent, and endstate info for tasks

jerrydevs:
AddTask and EditTask modals now contain 3 additional textarea fields to add additional info for tasks.

Requires API PR here: OneCommunityGlobal/HGNRest#27


Junhao6503 (approved):
looks good

PR #93, opened 2020-09-28
Changed the style of UserProfile and UserProfile; updated the form validation functionality for UserProfile Edit

Junhao6503:


chrisweilacker (commented):
Looks good so far.  Two things I noticed from running it. 
1. There isn't a way to switch back to the view profile after saving changes in the edit mode without hitting cancel, which isn't quite intuitive but can be added in a future update.
2.  When I switch something in the edit area from public to private it still shows up on my screen after saving and going back to the view area.  The Blue Squares section before also had the option to switch from public to private but dont see that option here. 

Also not sure what the More Tabs is,  I am guessing a place to add Projects and Team Memberships later.

src/actions/allTeamsAction.js
```diff
@@ -110,6 +110,7 @@ export const getAllUserTeams = () => {
   return async (dispatch) => {
     userTeamsPromise.then((res) => {
       dispatch(teamMembersFectchACtion(res.data));
+      console.log("getAllUserTeams: res:", res.data)
```
chrisweilacker, line 113: Think you should comment your console logs out.  Inspect has like at least 5 or 6 per page load.

src/actions/userProfile.js
```diff
@@ -11,6 +11,7 @@ export const getUserProfile = userId => {
 	const url = ENDPOINTS.USER_PROFILE(userId)
 	return async dispatch => {
 		const res = await axios.get(url)
+		console.log('GET user profile: response:', res)
```
chrisweilacker, line 14: Another console.log to comment out.

src/components/Teams/Teams.jsx
```diff
@@ -30,14 +30,14 @@ class Teams extends React.PureComponent {
...
       isActive: '',
-
     }
   }
 
   componentDidMount() {
     // Initiating the teams fetch action.
     this.props.getAllUserTeams();
     this.props.getAllUserProfile();
+    console.log('teams: props', this.props)
```
chrisweilacker, line 40: same comment out

chrisweilacker (approved):
Looks good to me.  If you aren't using them anymore can you find and comment out the console. Logs.  There seems to be a lot of them.

90756f45156e5d7d543134d1cd949e5

chrisweilacker (approved):
Looks good.

PR #102, opened 2020-11-22
Wbs4.2.1/reset password login and create new user profile

getnithesh:


chrisweilacker (approved):
I am approving this as we are trying to launch soon and nothing breaks existing functionality.

1. The forgot password link doesn't seem to go past the first page after I enter the email, fist name, and last name it just stays there. Please fix this.

2. After creating a new user if you want to change the password for some reason the first time I click the link it reloads the whole page instead of creating the pop up, after I go back and click change password again the popup does show.

3. After creating a new user and logging in with that user the profile image is listed as undefined.
691cb4c5c8f8a35caba069a19afce67

PR #105, opened 2020-11-29
badge section

zhangwen829:


zhangwen829:
For now data of badges was added to the public-use admin account, so logging in with other accounts won't see any badges. To check the badge display, please log in with "devadmin@hgn.net". The credentials for the admin account are included in the Instructions for Running HGN React App locally.

chrisweilacker (approved):
Looks good images/text needs work but will leave that up to jae to figure out. Good job.

PR #107, opened 2020-12-07
change to fix imageUrl null error

zhangwen829:
I changed some code, not sure if it will actually fix the crush happened on the dev preview site. But this fixed TypeError: cannot read property "imageUrl" of null on my local.


zhangwen829 (commented):
looks good.

chrisweilacker (approved):
Looks good.

PR #117, opened 2021-01-20
Fixes on multiple components

chrisweilacker:
Removed Progressbar on Timelog and added in Summary Bar
Fixed Style Warning and errors in console
Modified Progress Bar on Leaderboard to show incremental changes
Fixed Timer to Clear once a successful timelog submit was invoked.
Fixed Rounded Values on Leaderboard
Allowed Leaderboard Data to be Auto Refreshed
Fixed TimeEntry Form to disallow Multi-Submissions


daoquangvy:
About the commits that related to "Fixed Rounded Values and MultSubmit on TimeEntry Form". I'm not sure if the way we currently rounded time is intended or not. Since If I work for 1 hour and 27 minutes, I will expect my work hour be rounded to 1 hour and 30 minutes instead of 1 hour and 45 minutes.  Also If I work for 2 hour and 30 minutes, the leaderboard show 1.95 instead of 2.30
image

I view review more tomorrow and edit this comment.

chrisweilacker:
> About the commit that related to "Fixed Rounded Values and MultSubmit on TimeEntry Form". I'm not sure if the way we currently rounded time is intended or not. Since If I work for 1 hour and 27 minutes, I will expect my work hour be rounded to 1 hour and 30 minutes instead of 1 hour and 45 minutes. Also If I work for 2 hour and 30 minutes, the leaderboard show 1.95 instead of 2.30
> image
> 
> About the commit that related to "Fixed style warnings, and errors in console as well as removed progre…" I also fix a same warning in few differnet file when we try to catch 401 error, or try to compare
> 
> The rest of the commits. I don't see any things wrong with all the code.

1 hour and 27 minutes is 1.45 hours, not 1 hour and 45 min... because it is a . not a :  that means it's a fraction and 30 minutes is half an hour or .5 hours so 27 minutes I guess comes out to .45.   

I don't see this in the screenshot:   Also If I work for 2 hours and 30 minutes, the leaderboard show 1.95 instead of 2.30

One thing to make sure is all the time is logged with tangible time when you manually add the time you have to click the checkbox I think.   If it isn't tangible then the first 1.45 will not be updated but the second total time in your line should be updated.

Also if it all looks good you should do a review and click approve. Thanks.

daoquangvy (approved):
Sorry for the misunderstanding.
This is the image that I attached above
https://i.imgur.com/cwlKm34.png
I thought we use the normal format hour: minute (60 minutes), so that's why I think we should round up from 1.27  to 1.30, or round-up from 1.97 to 2.30. However, It makes sense for me now that we use the decimal time and the round-up is correct.

PR #118, opened 2021-01-30
Wbs4.2.1- Create/Edit  user profile changes

getnithesh:


daoquangvy (approved):
I'm approving this. These are few comments that I have for this PR:

1. I'm able to change update my password on the UserProfile page. I don't see any problems or bugs so far. However, I wonder if we should consider asking the user to enter their current password, and also log them out of the account after they change their password

2. When I created a new account, it does use the new default password that you change "123Welocme!". So, It works perfectly.

PR #119, opened 2021-01-31
Fix/Updated unit-testing files.

daoquangvy:
Fix few unit-testing files. I will work on more of them.


src/__tests__/mockStates.js
```diff
@@ -422,7 +422,7 @@ export const allProjectsMock = {
 };
 
 export const authMock = {
-  isAuthenticated: true,
+  isAdmin: true,
```
chrisweilacker, line 425: This looks good to add, but looks like you deleted isAuthenticated: true 
I am not sure if this is what you wanted, if this will error out on anything else as this was Jun's code so I don't know what it was needed for.  I will go ahead and merge, but wanted to point this out.  Also the 83% I am not sure where you see that code coverage can you let me know as it shows 45% for me.  Thanks.

chrisweilacker (approved):
The 83% I am not sure where you see that code coverage can you let me know as it shows 45% for me.  Thanks.

src/__tests__/mockStates.js
daoquangvy: 1. I change the isAuthenticated from the mock state to isAdmin because the Timelog.jsx was using "isAdmin" rather than "isAuthenticated". However, I can add it back in the future, If other component use isAuthenticated.
2. I run "yarn test --coverage", and it says that the unit-testing cover 83%
image

PR #122, opened 2021-02-06
Improved speed and added notification to Timers

chrisweilacker:
Improved speed by reversing warning fixes that caused network errors 
Added notification to Timer


daoquangvy (approved):
1. The two-timer worked. It's a little bit glitchy since sometimes the time does not since. Sometimes, the timer will just start at a random time that the browser remembers when I stop before I start it again. If I can give some suggestions, I wonder if there is a way that after we can make the browser remember when the user clicks, and it will keep continue counting the time even if the user closes the browser. That way, if the user opens the browser in a different tab the time is still counting, and they can only pause or stop the time rather than start a different one.
2. Everything else seems to work normally. I don't see any problems.

I will approve this and merge it.

PR #124, opened 2021-02-14
End date in user profile.

getnithesh:


daoquangvy (approved):
1. Everything works normally, Pause button also works and it updates the under Paused until column.
2.  I don't see any problems, bugs, so I will approve this.

PR #137, opened 2021-03-27
fix : corrections and message to remember the user to save the changes inside the user profile

getnithesh:


daoquangvy (approved):
The message does show up when the user changes their user information. I will approve this since it's work.

However, I wonder if you can make sure that the "Alert" disappear after user submit their change @nithuan141  ? Becuase right now, it won't disappear after we submit the change.

PR #176, opened 2021-07-31
update to summary bar

NickTSA:
summary bar links to route "/weeklysummaryreport". color change for "!" to red. Name has a break on large display and is one line on all other display sizes.


chrisweilacker (dismissed):
lgtm

PR #178, opened 2021-08-02
Blue square changes

CameronWhiteCS:
From medium severity bug 2A:

> When an Admin edits a blue square, clicking the “Update” or “Delete” button should be the only action needed to update it. Right now it doesn’t save to the DB, so I have to update something else to make the “Save” button active. Then my blue square edits will be saved along with whatever other edit I made

Summary: 
- When an admin modifies a user's blue squares, modifications to that user's profile are automatically submitted. This includes changes to the user's profile that have nothing to do with blue squares.


src/components/LeaderBoard/LeaderBoard.container.jsx
```diff
@@ -39,12 +39,12 @@ const mapStateToProps = state => {
...
 			// organization.intangibletime += _.round(element.intangibletime, 2);
-			
+
 			return element
 		})
 	}
-	
-	orgData.name = `HGN Totals: ${orgData.member_count} Members`
+
+	orgData.name = `HGN Totals: ${leaderBoardData.length} Members`
```
chrisweilacker, line 47: This won't work in all cases I will fix it myself though.

chrisweilacker (approved):
There is still a bug when you add a blue square then try to modify it afterwards without refreshing, we should add to the bug list but overall lgtm.

chrisweilacker (approved):
lgtm

PR #180, opened 2021-08-03
Low priority 5a: timer stops counting after clicking stop

CameronWhiteCS:
Fixes low priority bug 5a.

> The time log should keep running until "submit" is hit. This way people get credit for the time it takes them to write their entry and/or they can use that open window to keep a record of what they are doing, while the timer keeps running 

Summary: 

- The timer will now continue to count up until the entry has been submitted, not after the stop button has been pressed. This makes time log entries take into account how long it takes to fill out the form as intended.


chrisweilacker (approved):
lgtm

PR #181, opened 2021-08-04
High priority 3A and 3B: Fixed issues resulting from differences in local time versus PST

CameronWhiteCS:
High priority 3A and 3B: Fixed issues resulting from differences in local time versus PST

> Entries from the previous week with correct (last week’s) timestamp and descriptions appear in the current week (see 5 hours 6 min in first image). Good news is, in the list of team members it shows the correct time for the current week (0 hours in the second image)

> There’s also an issue with people in other timezones logging time just a few hours before the week closes Saturday night PDT and those hours not counting for their current week… they get logged in the next week. So it seems the timer isn’t 100% synced with Pacific Time.

Summary:

-  When displaying time entries, local time is converted to PST before the times are fetched. So, if you roll over into the next week in your time zone but it’s still last week in PST, you won’t see the next week’s entries until it’s also the next week in PST.

-  When submitting a new time entry, local time is converted to PST for the default date value. If you’re in the next day in your time zone but it’s still yesterday in PST, yesterday is the default date to make sure it lines up with PST.

- The text “Viewing time Entries from xxx-xx-xx to xxxx-xx-xx” on the time log will only roll over to the next week once the next week begins in PST. Previously, local time was used

- The time entries you see are now dependent upon PST, not local time

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "jwt-decode": "^2.2.0",
     "lodash": "^4.17.21",
     "moment": "^2.25.3",
-    "moment-timezone": "^0.5.31",
+    "moment-timezone": "^0.5.33",
     "pdfmake": "^0.1.65",
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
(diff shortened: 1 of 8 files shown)


chrisweilacker (approved):
lgtm

PR #182, opened 2021-08-04
Time entry form no longer closes when pressing the clear button

CameronWhiteCS:


CameronWhiteCS:
Accidentally included in PR #183 making this PR redundant

PR #183, opened 2021-08-04
Time warnings

CameronWhiteCS:
Fixes medium priority bug # 7:

> Admin editing a person’s time should not contribute to their time-edit count towards blue squares. I did this on Malathi and got an email notice. The email notice is good, but I’d like it not to contribute to their blue square count… and for the message to say that it doesn’t contribute to it. 

Summary:

- Admins are no longer warned about editing time entries
- The edit count of time entries is not increased when edited by an admin
- Fixed bug allowing non-admins to change the tangibility of their time entries (PR #183)
- Warnings are no longer issued when editing **intangible** time, nor does editing intangible time increment the edit count.
- Pressing "Clear Form" when creating or editing a time entry no longer closes the menu due to a truthy type check (accidentally included from PR #182)

### DIFF
--- a/src/components/Timelog/TimeEntry.jsx
+++ b/src/components/Timelog/TimeEntry.jsx
@@ -25,8 +25,11 @@ const TimeEntry = ({ data, displayYear, userProfile }) => {
   const dispatch = useDispatch();
 
   const toggleTangability = () => {
-    
-    const newData = {...data,
+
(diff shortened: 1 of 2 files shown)


CameronWhiteCS:
Migrating changes from this PR into a refactor of the TimeEntryForm component as there was significant overlap

PR #185, opened 2021-08-05
Low priority 1a: Search UX

CameronWhiteCS:
Solves low priority 1a: 

> This should auto-populate the list with all active options. Right now you have to type a name and hit “enter”, which is way less efficient and user friendly.  

Summary: 

- Magnifying glass search button removed as it was made redundant
- Search results now auto-populate as you type


chrisweilacker (approved):
lgtm

PR #190, opened 2021-08-06
Time entry form refactor

CameronWhiteCS:
# Refactor of TimeEntryForm

## Description:

Refactored the TimeEntryForm component fixing several bugs and adding many requested features in the process.

Changes introduced in PR #183 have been included due to significant overlap and #183 has been closed. 

Fixes high priority #5, medium priority bug # 7A, and low priority #8

> If your internet connection is slow or unstable when submitting a tangible time entry, there’s a chance the timer will be reset but the submission will fail

> Admin editing a person’s time should not contribute to their time-edit count towards blue squares. I did this on Malathi and got an email notice. The email notice is good, but I’d like it not to contribute to their blue square count… and for the message to say that it doesn’t contribute to it.

> All user classes should be able to edit their time BUT IT MUST STILL GENERATE THE WARNING POPUP AND AUTOMATED BLUE SQUARE SYSTEM. This seems to work for Admins right now. (This absolutely can’t be integrated if it allows people to edit their time without tracking those edits and issuing blue squares for too many. So please wait to do this one until after we’ve shut down the Beta and are testing things like this on the Dev environment before pushing changes to the live environment.)

## Changes: 

- Updated test suites to account for new behavior

- The user is now notified if their time entry submission fails and the form remains open without being cleared. This is useful in the event of a loss of internet or other submission failure (see high priority #5)

- If a user's time entry is submitted successfully but there's a failure when asking the server to reset their timer, the user is notified and instructed not to submit the same time entry again. 

- Admins are no longer warned about editing time entries

- The edit count of time entries is not increased when edited by an admin

- Fixed bug allowing non-admins to change the tangibility of their time entries (PR #183)

- Warnings are no longer issued when editing **intangible** time, nor does editing intangible time increment the edit count.

- Non-admins can edit their own time entries as long as they were submitted the same day they are being editing (low priority # 7)***

- Pressing "Clear Form" when creating or editing a time entry no longer closes the menu due to a truthy type check

*** The functionality for this change has been implemented and can be introduced by uncommenting a single line. As stated above, we are waiting until we're no longer in the beta environment ad can issue automatic blue squares before this change is deployed.


YueruZhao (approved):
Looks good to me, please merge

PR #194, opened 2021-08-09
Name spacing fix

NickTSA:
Name spacing is fixed to fit longer names. container has been adjusted for better use of space and centered correctly. badges have been adjusted to be centered.

Improvements to code... simplified.


YueruZhao (approved):
Looks good, please merge

PR #196, opened 2021-08-10
Fixed tangibility problem

CameronWhiteCS:


one-community (approved):
Merging without review because change was simple

PR #199, opened 2021-08-11
Timer now visible on all screen sizes

CameronWhiteCS:
Nav bar branding removed and replaced with the timer. The timer is now also visible on every screen size and doesn't become smaller as the screen shrinks.


YueruZhao (approved):
Looks good, please merge

PR #202, opened 2021-08-11
Fixed bug causing inaccurate projects/tasks to be displayed

CameronWhiteCS:
Bug fix:
> Jae: 8/6: Still brings up another person’s projects list for me if I’m viewing their profile or timelog and try to submit my time using Timer → stop So I can’t submit my time unless I navigate back to my own dashboard. See pics below. Additional comment: “My mistake reporting this bug. I took pics 1 and 2 below (with Developers Tools open) to show you how I thought it still existed because you can see your info in the background and it only offering me YOUR projects to log time… Then I thought, "maybe it still logs to me" and it did! It swaps out your time log for mine when I log... but I could only log time to your project options, not my complete list... and part of the screen is still your information. See picture 3…. See how it says "Add time entry for Cameron", and your name top left, but then the time log below is actually mine. So is the time bar by your name... but the "!" info is still relevant to you and not me.”


YueruZhao (approved):
Looks good, please merge

PR #216, opened 2021-08-18
'Clear Form' now behaves as expected on TimeEntryForm

CameronWhiteCS:
- 'Clear Form' now behaves as expected on TimeEntryForm.
- If submitting multiple intangible time entries, the form properly clears itself in between submissions.


NickTSA (approved):
👍

PR #223, opened 2021-08-20
Fixed bug preventing the editing of weekly summaries

CameronWhiteCS:
Fixed high priority bug: 

> If a person saves their summary, and then goes back in and tries to add something to it, their new additions won’t save. It says it saves, but then you return to the page and the new additions aren’t there. Not sure if this bug means a person can’t save their summary the first time or not. This is a much bigger problem if that’s the case


YueruZhao (approved):
Looks good, please merge

PR #224, opened 2021-08-20
Fixed leaderboard displaying time improperly

CameronWhiteCS:
Solves issue:

> Dashboard → Leaderboard → Looking at the HGN Totals (see below)
This is not totalling the amount of committed time (1925.8 in the pic below) correctly. This total should be a sum of all active members’ weekly “tangible time” committed.


YueruZhao (dismissed):
Looks good, please merge

one-community (approved):

PR #231, opened 2021-08-24
Fixed bugs resulting from submitting time entries while in other time…

CameronWhiteCS:
Fixes issues :
>  Timer Functionality
Something seems to be off with the timer again. I got the reports below from people in other timezones. 
* Fixes a bug where time entry forms are not cleared after submission.


one-community (approved):
Approving without checking, as per Cameron's recommendation

PR #259, opened 2021-08-28
Implemeted auto reload on update feature

CameronWhiteCS:
# HGN Application now auto-reloads on update

The client is now able to detect that it's been updated and inform the user of the update (auto refresh could cause problems with erasing unsaved work). This document describes how this process works. 

## Mechanism

1. During the CircleCI build process, the build directory is compressed into a zip archive and the sha256 hash of the .zip file is is calculated. The resulting hash is then stored inside of the build/hash.txt file. 

2. When the client initializes, it requests a copy of hash.txt and stores its contents in memory.

3. After initialization, the client polls the server every 5 minutes and asks for /hash.txt. If the contents match, nothing happens. If they differ, the client informs the user of the update.

4. This means that each time a new version of the client is deployed, a new, different hash.txt file will be produced automatically. 

## Motive 

This approach is better than using a version string as it is fully automated and it isn't possible to forget to update the version string. 

## Development Environments

To ensure that this feature doesn't cause issues while working in a development environment and to provide a failsafe mechanism, a default hash.txt file has been added to the project's public folder. Because the hash is generated during deployment, running the app locally will not interfere with this file.

### DIFF
--- a/src/components/AutoUpdate/AutoUpdate.jsx
+++ b/src/components/AutoUpdate/AutoUpdate.jsx
@@ -1,0 +1,66 @@
+import { Container, Row, Col, Alert } from 'reactstrap'
+import { useState, useEffect } from 'react'
+
+const AutoUpdate = () => {
+  const SECOND = 1000
+  const MINUTE = 60 * SECOND
+
+  const [updated, setUpdated] = useState(false)
+  const [hash, setHash] = useState(undefined)
+
+  const noCacheHeaders = new Headers()
+  noCacheHeaders.append('pragma', 'no-cache')
+  noCacheHeaders.append('cache-control', 'no-cache')
+
+  const requestParams = {
+    method: 'GET',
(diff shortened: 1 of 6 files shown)


src/components/AutoUpdate/AutoUpdate.jsx
```diff
@@ -0,0 +1,66 @@
...
+            response.text().then(text => {
+              if (text !== hash) {
+                setUpdated(true)
+              }
+            })
+          })
+          .catch(err => {})
+      }, 5 * MINUTE)
+
+      return () => clearInterval(interval)
```
calvertwong, line 46: Is clearInterval necessary? Won't it stop the process?

calvertwong: My bad, this is needed. All good now.

calvertwong (approved):
Interesting, approved.

PR #297, opened 2021-09-05
Leaderboard centering of data and active green icon sizing

NickTSA:
Jae: Dashboard → Leaderboard also Admin View → Profile Pages (WIP: Narek)
Admin View → Profile Pages: Please increase the size of the green dot to match the size of the other icons here
Screen Shot 2021-09-05 at 12 30 35 AM

Dashboard → Leaderboard: This will make it bigger on the Dashboard too. Right now it is higher than the words there, please also center it vertically (like the name there). Please vertically center everything else in the rows there too. You can see below how only the name is centered vertically in the picture below. Everything else should be centered like that too.
Screen Shot 2021-09-05 at 12 30 49 AM


calvertwong:
Could you attach some screenshots?

CameronWhiteCS (approved):

PR #298, opened 2021-09-07
Heavily refactored the UserProfile component

CameronWhiteCS:
# Heavy refactoring of the UserProfile component

## Motivation

Recently, there have been a large number of bugs affecting the UserProfile component that are difficult to reproduce on demand in a testing environment and are highly disruptive. This PR attempts to fix these issues by redesigning how a user profile object is stored in memory when using the UserProfile component. Additionally, some code cleanup was also preformed. 

## Summary of changes

- The monolithic handleUserProfile function has been almost entirely eliminated. It previously managed state changes for 15 - 20 input fields and now handles only 3. 

- When viewing a user profile, the global redux store is no longer modified

- The user profile you're viewing no longer has to be the same one that's stored in the global store. Previously, only one user profile could be loaded in memory at a time, which caused many issues and hard to reproduce bugs.

- The tests for UserProfile have been temporary disabled because the implementation of UserProfile has changed significantly.

### DIFF
--- a/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
+++ b/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
@@ -37,7 +37,6 @@ import PhoneInput from 'react-phone-input-2'
 import 'react-phone-input-2/lib/style.css'
 
 import classnames from 'classnames'
-import timeZoneMap from 'constants/timeZones'
 import TimeZoneDropDown from '../TimeZoneDropDown'
 
 class AddUserProfile extends Component {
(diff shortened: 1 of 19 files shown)


src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
```diff
@@ -244,7 +244,10 @@ class AddUserProfile extends Component {
                     <FormGroup>
                       <TimeZoneDropDown
                       filter={this.state.timeZoneFilter}
-                      onChange={this.handleUserProfile}
+                      onChange={(e) => {
+                        this.handleUserProfile(e);
+                        console.log('@@@@@@@@@@@@@@@@@@@@@@@@@@');
```
calvertwong, line 249: Remove console.log?

CameronWhiteCS: Fixed this issue and added a couple of changes I forgot to add originally.

calvertwong (approved):

src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
calvertwong: That's fine but there is one conflict. I believe it is because of the tiny addition of the endDate portion in UserProfile.js.

calvertwong (approved):

one-community:
Tried a ton of functions on this on the Dev and everything seams to be working great!

PR #299, opened 2021-09-07
Fix/display end date volunteering time tab

calvertwong:
Bug title: Jae: Profile → Volunteering Times → Start Date/End Date

Added an editable input in the volunteering time tab for admin to edit end date. The endDate field will be added to Mongo if the date is valid and when the user is paused at the Basic Information tab with current date as the endDate or at the Volunteering tab by adding an endDate.

image


CameronWhiteCS:
There have been issues with Jest when ```moment``` is used instead of ```moment-timezone```. Changed package to ```moment-timezone``` and approved.

CameronWhiteCS (approved):

PR #304, opened 2021-09-09
Made the authentication system more secure and follow established sta…

CameronWhiteCS:
# Authentication Mechanism Overhual

**WARNING**: Both the front-end PR and the back-end PR for these changes are breaking changes. All logged in users at the time of this update will likely recieve numerous errors. Additionally, both the front-end and back-end PR must be deployed at the same time to avoid problems. Corresponding back-end PR: https://github.com/OneCommunityGlobal/HGNRest/pull/98

## Motivation 

The server and client previously used a JWT-only authentication flow. The server signed a JWT with a lifespan of **10 days** and the client provided it every request until it expired. This is far beyond the recommended lifespan for a JWT which is usually only a few minutes (5 - 15), or upwards or 24 hours for more benign oAuth 2 use cases. 

The main issue with a JWT-only system is that user access to the system fundamentally cannot be revoked, at least not in a timely fashion. This is because in this kind of a system, a message that has been signed will always be valid unless the private key it was signed with changed. Until now, there was no secure way to respond to the following scenarios:

- User's permissions need to be downgraded
- User's account is hijacked (hijacker could of have access for up to 10 days)
- User needs to be blocked / banned from system
- User's old sessions need to be invalidated (like during a password reset)

The JWT-only system was replaced with a new system that uses a combination of JWT's and rotating refresh tokens as described in the next section.

## Solution

JWT's issued by the server now last for only 15 minutes (hard coded). The environment variables that previously described the duration of the JWT's have been repurposed and now describe the lifetime of refresh tokens. 

A refresh token is a **revokable** token that can be exchanged for both a new refresh token and a new access token (JWT). Each time a refresh token is used, it becomes invalid and the client is given a new refresh token to replace it. This scheme is referred to as a **rotating refresh token**.

Now, every 15 seconds, the client checks if it's access token will expire within the next 60 seconds. If that's the case, the client exchanges its refresh token for a new access token and a new refresh token.  

## Implications

- Access to the system can now be properly revoked

- Because of the rotating refresh token scheme, users never have to sign into the application again unless they remain inactive for more than 10 days. Each time a refresh token is used, the new token granted in response has a new expiration date 10 days in the future. 

## Technical Summary 

- Added 'Bearer' prefix to HTTP Authorization header

- Replaced non-standard 'expiryTimestamp' field of the JWT payload with the standard 'exp' field

- The JWT's issued by the server now only last 15 minutes and they are renewed using a refresh token.

- New authentication scheme is compatible with a an isolated authentication service

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,10 +1,11 @@
-import React, { Component } from "react";
+import React, { useState, useEffect } from "react";
 import routes from '../routes'
 import logger from "../services/logService";
 
 import httpService from "../services/httpService";
 import jwtDecode from 'jwt-decode';
 import { setCurrentUser, logoutUser } from "../actions/authActions"
+import moment from 'moment-timezone'
 
 import { Provider } from 'react-redux';
 import { BrowserRouter as Router } from 'react-router-dom';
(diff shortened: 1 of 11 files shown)


src/components/App.jsx
```diff
@@ -1,77 +1,80 @@
...
+    })
+    .then((res) => {
+      const decoded = jwtDecode(res.data.token);
+      httpService.setjwt(res.data.token);
+      store.dispatch(setCurrentUser(decoded));
+      localStorage.setItem('refreshToken', JSON.stringify(res.data.refreshToken))
+      setInitialized(true);
+    })
+    .catch((err) => {
+      console.log(err);
```
calvertwong, line 58: remove console.log

CameronWhiteCS: Fixed

calvertwong (dismissed):

chrisweilacker (approved):
lgtm lets test further in dev

PR #307, opened 2021-09-11
User management page sets endDate on deactivating volunteer

calvertwong:
Set endDate on activate deactivate button click

endDate field will be added to Mongo when the activate deactivate button/icon is clicked.

### DIFF
--- a/src/actions/userManagement.js
+++ b/src/actions/userManagement.js
@@ -8,7 +8,6 @@ import {
 } from '../constants/userManagement'
 import { ENDPOINTS } from '../utils/URL'
 import { UserStatus } from '../utils/enums'
-import moment from 'moment-timezone'
 
 /**
  * fetching all user profiles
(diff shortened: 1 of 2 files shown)


CameronWhiteCS (approved):

one-community:
This only partially worked for me. Inactive date was not shown here on the person's Profile Page: Screen Shot 2021-09-11 at 12 21 02 PM Also, making a user inactive by clicking the (now grey) dot there didn't cause the date to be added on the Team Management page or under the Profile --> Volunteer Times tab

I also see people on the Other Links --> User Management page with end dates listed even though they are still active:
Screen Shot 2021-09-11 at 12 28 05 PM If I scroll back to later pages in the list there (Other Links --> User Management --> (bottom of page) Previous --> 11) there are also people that are inactive that don't show an end date. 

This odd data also showed up under the Other Links --> User Management page but then converted to the correct date a minute or so later: 
Screen Shot 2021-09-11 at 12 20 19 PM Not sure but this may be due to me currently updating the back end.

PR #322, opened 2021-09-25
Add end date to db when click circle icon in user profile page

calvertwong:
Before
After

After the dot is clicked, end date appears and will be stored in the DB once the submit button at the bottom is clicked.

### DIFF
--- a/src/components/UserManagement/UserTableData.jsx
+++ b/src/components/UserManagement/UserTableData.jsx
@@ -4,7 +4,6 @@ import { DELETE, PAUSE, RESUME } from '../../languages/en/ui';
 import { UserStatus } from '../../utils/enums';
 import { useHistory } from 'react-router-dom';
 import ActiveCell from './ActiveCell';
-import moment from 'moment-timezone';
 
 /**
  * The body row of the user table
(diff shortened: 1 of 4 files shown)


src/components/UserProfile/VolunteeringTimeTab/VolunteeringTimeTab.jsx
```diff
@@ -225,8 +225,8 @@ const ViewTab = props => {
       </Row>
 
       {Object.keys(props.userProfile.hoursByCategory).map(key => (
-        <>
-          <Row key={'hours-by-category-' + key}>
+        <React.Fragment key={'hours-by-category-' + key}>
```
calvertwong, line 228: This line get rid of one warning about missing key.

src/components/UserManagement/UserTableData.jsx
```diff
@@ -61,7 +60,7 @@ const UserTableData = React.memo((props) => {
           ? (props.user.reactivationDate.toLocaleString().split('T')[0]) : ''}
       </td>
       <td>
-        {props.user.endDate ? (moment(props.user.endDate).format('YYYY-MM-DD')) : 'N/A'}
+        {props.user.endDate ? (props.user.endDate.toLocaleString().split('T')[0]) : 'N/A'}
```
hannahgardner96, line 63: When I look up the behavior of tolocalestring and test it in fiddle, I don't see any returns that include a T. What is the motivation for this change?

calvertwong: Are you thinking that this line of code is returning something with the 'T' in it? If yes, that is not true. What I am doing here is to take a portion of the endDate which will look something like this `2021-09-30T00:00:00.000+00:00`, for this case it will be `2021-09-30`

The reason I am using toLocalString() instead of toString() is just to match the code in line 60 so that the code is more neat.

hannahgardner96: Not whether the line of code returned something with a T, just the utility of the split("T") as it's used with toLocaleString but you answered my question!

hannahgardner96 (approved):

PR #335, opened 2021-11-13
form validation occurs onfocus and typo fix

hannahgardner96:
Other Links → User Management → Create New User
Mandatory fields should be indicated and a popup should happen that says “Please complete all mandatory fields” if they aren’t. 

note: form validation with error messages was already present but only occurred onChange and onSubmit. I added it to occur onFocus so that a user cannot enter and exit a required field without a notification.


navyasrim (commented):
Checked the code..no conflicts as per my understanding it was not blocking any where..

one-community:
Did you confirm/test also that it works as it should? If so, approve it but don’t merge it.

navyasrim (approved):
Approved as the code is not blocking any where--Navya.

PR #341, opened 2021-12-18
Can’t delete a badge from a person’s Profile: Profile → Select Featu…ed → Delete → Save Fixed this Issue

navyasrim:
Fixed the Issue 1 (Can’t delete a badge from a person’s Profile: Profile → Select Featured → Delete → Save )
Can’t delete a badge from a person’s Profile: Profile → Select Featured → Delete → Save Changes. When I do this, nothing happens. 
Still works from the person’s Dashboard though, so Chris said it must be some change to the Profile Page that broke this function that was working before. 
Hitting “save” doesn’t close the window: If I then click the “x” to leave the window, I get a blank screen
Can’t edit/choose featured badges. I did it once, then it wouldn’t let me do it again and save. Profile → Featured Badges → Select Featured → choose the ones I want featured → Save Changes
------
The changes i have modified in these **pages:**
components/userprofile/Badges.jsx
Modified in line 39 ,commented the previous code and added modified code by passing correct userId props
changed the props previously user id is not passing now i changed with correct props
userId={props.userProfile._id} 

Also changed in Components/badge/BadgeReport.jsx
added this code to reload the page after closing or clicking on save and closing the modal.
//Reload the view profile page with updated bages
  window.location.reload();

### DIFF
--- a/src/components/Badge/BadgeReport.jsx
+++ b/src/components/Badge/BadgeReport.jsx
@@ -218,6 +218,9 @@ const BadgeReport = (props) => {
     await props.getUserProfile(props.userId);
     //close the modal
     props.close();
+        //Reload the view profile page with updated bages
+        window.location.reload();
+
   }
 
   return (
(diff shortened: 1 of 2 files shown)


rachit-joshi (commented):
Any reason you have kept the old code commented there? Might be better to get rid of it to maintain cleanliness.

rachit-joshi (approved):

navyasrim:
> Any reason you have kept the old code commented there ( line 38 in components/userprofile/Badges.jsx)? Might be better to get rid of it to maintain cleanliness. Looks good other than that.

No reason...as I just kept it so that in future if we want we can uncomment out...

navyasrim:
reopoened this PR for merge.

PR #352, opened 2022-01-23
Issue4 fixed, show mandatory fields that aren’t completed when a person click Create button

navyasrim:
Issue 4: Mandatory fields should be indicated and a popup should happen that says “Please complete all mandatory fields” if they aren’t. Currently they only do this if a person clicks in a field and doesn’t enter anything. They should show mandatory fields that aren’t completed when a person clicks “Create”

Resolved this issue by removing OnFocus method and calling or checking this in Button click event.
Now the mandatory fields are showing in the toast message.


navyasrim:
Will do new PR

PR #354, opened 2022-01-23
Issue4 Required fields are now showing on click create button

navyasrim:
Issue 4 : Mandatory fields should be indicated and a popup should happen that says “Please complete all mandatory fields” if they aren’t. Currently they only do this if a person clicks in a field and doesn’t enter anything. They should show mandatory fields that aren’t completed when a person clicks “Create”

Fixed this issue by removing the OnFocus and implemented the logic in Create button click event. The required fields will also show in toast message.
wk12_userprofiletesting

### DIFF
--- a/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
+++ b/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
@@ -432,8 +432,6 @@ class AddUserProfile extends Component {
   }
 
   createUserProfile = () => {
-    this.setState({ formSubmitted: true })
-
     let that = this
     const {
       firstName,
@@ -466,6 +464,8 @@ class AddUserProfile extends Component {
       collaborationPreference: collaborationPreference,
       timeZone,
     }
+
+    this.setState({ formSubmitted: true })
 
     if (googleDoc) {
       userData.adminLinks.push({ Name: 'Google Doc', Link: googleDoc })
--- a/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
+++ b/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
@@ -40,7 +40,7 @@ import 'react-phone-input-2/lib/style.css'
 
 import classnames from 'classnames'
 import TimeZoneDropDown from '../TimeZoneDropDown'
-var patt = ''
+const patt = RegExp(/^([\w.%+-]+)@([\w-]+\.)+([\w]{2,})$/i)
 class AddUserProfile extends Component {
   constructor(props) {
     super(props)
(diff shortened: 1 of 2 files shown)


aaronleechan:
navyasrim  : 
Please remove the hard coding message from <FormFeedback>{XXXXXX}</FormFeedback>
Please take the error message from this.state.formErrors
You just need to pop up the highlight and error message from the state.

Phone number is also not working properly after user fill up the number.

src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
```diff
@@ -430,14 +436,25 @@ class AddUserProfile extends Component {
     if (googleDoc) {
       userData.adminLinks.push({ Name: 'Google Doc', Link: googleDoc })
     }
-
-    createUser(userData)
+    if(this.state.userProfile.phoneNumber !== null)
```
getnithesh, line 439: can we move all these conditions to a function and use that function here before calling createUser?

src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
```diff
@@ -40,7 +40,7 @@ import 'react-phone-input-2/lib/style.css'
 
 import classnames from 'classnames'
 import TimeZoneDropDown from '../TimeZoneDropDown'
-
+var patt = '';
```
ireneclaregs, line 43: I dont see any reference of this global variable and if you really need, Can you please update it to "let pattern"?

src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
klbeg: conditions moved to:  inputsHaveLength()

src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
klbeg: changed 'part' to a const and removed duplicate occurrences

klbeg:
@aaronleechan hard coding of errors is needed if the input fields haven't been focused, but if they have been, formErrors are now displayed instead.

Errors now displayed if users don't focus any fields

'Phone number required" message now appears even if the user adds and deletes a number

Toast error of  "please complete all mandatory fields" now shows if fields are left blank

one-community:
Aaron has wrapped up his work with us. So let me know if this is ready and I'll ask Irene to review it.

klbeg:
Ok.  I'll finalize tomorrow.  I want to double check the code I pushed this morning. 

Thank you!

one-community:
Sounds good, thank you.

ireneclaregs (approved):
LGTM

PR #357, opened 2022-01-30
update delete badge message

aaronleechan:
Other Links → Badge Management → Assign Badge → Confirm
Doing the above should be all that is needed. Currently it then requires a person to hit “Submit” next for changes to take place. This is confusing and inefficient. I’d like “Confirm” to be all that is needed. 

Update the Message as require.


getnithesh (approved):

ireneclaregs (approved):
LGTM

PR #363, opened 2022-02-24
Klbeg delete badge pop up

klbeg:
Changed text on confirmation popup to requested text with line break.  

Also changed the window.alert for 'can only select 5 badges at a time' to a toast error so it's consistent with other error messages.


ireneclaregs (approved):
LGTM

PR #377, opened 2022-03-18
Elyse

elyselam:
Bug: "3)Jae 1/4 - problem experienced again by me and another manager: Timer Logging Time to Wrong Day (WIP Elyse)
This bug is so odd and one I just experienced for the first time myself 10/18. The same person has reported it for a couple weeks in a row now (see below). Note it logged for 10/10, not 9/10 as she said, so that’s a mistake in how she described it. 
My experience is that I just logged time using the timer and the time logged for yesterday. Not positive, but these time errors seem to always backlog to the Sunday prior to whatever date the person is trying to log to. 
Both myself and the person communicating below have also had the same experience of this bug happening, then immediately NOT happening for a time entry made right after the one where the bug happened. So it seems to fix itself after the first error. "

Fix:
I found the bug was caused by TimeEntryForm being rendered right when the page loaded. This is what populates the date to log the time user signed in instead of when they actually clicked the Stop button. I wrapped the TimeEntryForm in the condition for only when Modal is true, so the form pops up and date is populated with the current date instead of date user logged in.
I also reformatted with Prettier, cleaned up and removed unused console.logs.


elyselam:
When i set up VScode prettier, it made changes the these files. Do i need to .gitignore them before my pull request?
.eslintrc.js
package-lock.json
package.json

migfernandes01 (approved):
Everything looks good, passed my tests. Great job

PR #383, opened 2022-03-23
Team member tasks refactor

chrisweilacker:


src/components/TeamMemberTasks/TeamMemberTasks.jsx
```diff
@@ -15,257 +19,223 @@ import { getUserProfile } from '../../actions/userProfile';
...
+              .startOf('week')
+              .subtract(0, 'weeks');
+        
+            const toDate = moment()
+              .tz('America/Los_Angeles')
+              .endOf('week')
+              .subtract(0, 'weeks');
+        
+            memberTimeEntriesPromises.push(
+              httpService.get(ENDPOINTS.TIME_ENTRIES_PERIOD(member._id, fromDate, toDate)).catch((err) => { }),
```
chrisweilacker, line 124: For Every Member we are calling this endpoint, we need an endpoint that takes the list of members and returns the data with just one call to mongo.  I mean why are we even getting this data in team member tasks we should just be getting aggregated data where the total hours are calculated for you.

src/components/TeamMemberTasks/TeamMemberTasks.jsx
```diff
@@ -15,257 +19,223 @@ import { getUserProfile } from '../../actions/userProfile';
...
+                timeEntries: data[i].data,
+              };
+            }
 
-              // calculate hours done in current week and add to user obj for ease of access
+            console.log('members: ', uniqueMembers);
+        
+            // fetch all tasks for each member
+            uniqueMembers.forEach((member) => {
+              teamMemberTasksPromises.push(httpService.get(ENDPOINTS.TASKS_BY_USERID(member._id)).catch((err) => { if (err.status !== 401) { console.log(err); } }));
```
chrisweilacker, line 142: We should call one endpoint send that endpoint the list of user_id's and that endpoint should return back the tasks for those user ids with one call to mongo, not a call to mongo for each and every member.

migfernandes01:
Refactored calls to the Database, so instead of calling the same endpoints multiple times, now it calls the endpoint one time for a list of users.

EvianTan (approved):

PR #384, opened 2022-03-31
Added logout modal

migfernandes01:
Welcome, Dev (dropdown menu) > Logout 
Clicking cancel in the dialog leaves the user on a blank page instead of the page they were on when they attempted to log out.
This dialog should be changed from window.confirm() to a reactstrap Modal.  Use the confirmation from:  view profile > select featured (badges) > delete (any badge), as a template.

### DIFF
--- a/src/__tests__/ForcePasswordUpdate.test.js
+++ b/src/__tests__/ForcePasswordUpdate.test.js
@@ -36,7 +36,10 @@ const server = setupServer(
   rest.get('http://localhost:4500/api/userprofile/*', (req, res, ctx) => {
     return res(ctx.status(200), ctx.json({}));
   }),
-  rest.get('http://localhost:4500/hash.txt', (req, res, ctx) => {
+  rest.get('http://*/hash.txt', (req, res, ctx) => {
+    return res(ctx.status(200), ctx.json({}));
+  }),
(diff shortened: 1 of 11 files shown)


EvianTan:
I checked the change in this PR, local build succussed, and I verified the behavior of "click the logout button", it worked well. However, there is a build failure on `UnAuthenticated Header page` I think you need to fix...

migfernandes01:
@EvianTan Thanks for the comment, I will try to solve that problem.

chrisweilacker (approved):
lgtm, but get others to look at it too

ronBP95:
Taking a look at this to provide a second set of eyes for confirmation

ronBP95:
The logout modal looks fantastic and all of the buttons and redirects work as intended.
- Clicking the "X" button closes the modal
- Clicking the Cancel button closes the modal and the issue with a blank page redirect has been resolved
- Clicking logout redirects the user to the login page

ronBP95:
This appears to be unrelated but for some reason in this branch, there is an error that is coming up that is related to team member tasks.

×
Unhandled Rejection (TypeError): Cannot read properties of undefined (reading 'data')
/HighestGoodNetworkApp/src/components/TeamMemberTasks/TeamMemberTasks.jsx:150

PR #385, opened 2022-04-02
Remove "Search For Time Zone" feature in BasicInfo

elyselam:
Bug:
Profile Page → Basic Information
In addition to the above request, the “Search for Time Zone” field should be removed because the above request would replace this.


ronBP95 (commented):
The lines that were commented out are going to be removed in an upcoming change.

thienphuvn95 (approved):

PR #394, opened 2022-04-16
fix dashboard error

nchen0:
- Fix error on dashboard where data is not found. Added checks to not resolve the promises if there's nothing to resolve.
- line 115 & 121 are the only changed, github sees so many lines changed because the if statement pushed everything underneath one space in.


migfernandes01 (approved):
Passed all my test and code looks good. Since I worked for a while in this component recently, I already knew the logic and that was a good catch and fix

PR #398, opened 2022-04-20
fix - "Cannot read properties of undefined"

elyselam:
Fix - Gets error "Unhandled Rejection (TypeError): Cannot read properties of undefined (reading 'data')" in TeamMemberTasks.jsx when git pull.
Already tried these options in both back and front-end but get the same error:
git fetch origin development
git reset --hard FETCH_HEAD
git pull
npm install
npm run start:local


EvianTan (dismissed):

migfernandes01 (approved):
Passed all my tests and looks good. Great job!

PR #403, opened 2022-05-01
fix timezone and location persistence on user add

nchen0:


stevenwang213 (approved):
The following function was tested:

- Create a new user with the location specified. 
- Change location and timezone in the basic information tab and click "save", both fields are updated immediately.

This function works as expected! 👍

EvianTan (approved):

PR #405, opened 2022-05-06
Yiyun create timezone diff func

EvianTan:
Create the ability for user to intuitionally see the timezone difference between his own and the person he's viewing.
Note: 
- When the viewer views other's profile, it shows the timezone difference
- When the viewer views his own profile, it shows "you're viewing your own profile page"
- Cover the corner case of daylight saving by accessing today's date


stevenwang213 (approved):
The following functions were tested:
- View other's profile, it shows the correct time difference including cases without daylight saving time zone.
- View self profile, it shows the correct message.
Great job!

PR #413, opened 2022-05-18
Fixed Wrapping issue by adding a media query + overscroll-x property …

ronBP95:
…when page size shrinks to a certain point.

The following changes now trigger a scrollbar when the page would normally wrap. Please see the attached image below.

overflow-x


stevenwang213 (approved):
Shrink the page size and the badges block change to scrollable. Work as expected! Great Job!

PR #420, opened 2022-05-25
Modified CSS to repair the issue with wrapping

ronBP95:
Changes:

- Corrected issue with flex property for .badge_featured_container in badge.css (was previously flex: none -> is now flex: nowrap)
- Removed flex property from media query. This is likely the reason that there was a conflict.

### DIFF
--- a/src/components/UserProfile/Badge.css
+++ b/src/components/UserProfile/Badge.css
@@ -88,7 +88,7 @@
 
 .badge_featured_container {
   display: flex;
-  flex-wrap: none;
+  flex-wrap: nowrap;
   overflow-y: none;
 }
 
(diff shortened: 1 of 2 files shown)


ronBP95:
I also modified the media query type to "screen and", which was not previously specified before. This may have been causing some browsers to ignore the media query.

stevenwang213 (commented):
Test on several users w/wo badges and adjust the window size. Everything works well. Great Job!
image

stevenwang213 (approved):

ronBP95:
@stevenwang213 I wasn't able to replicate the "Assign Badges" button being clipped in the image. After doing some investigating however, it looks like the width of the cardTitle child element exceeds the width of the parent element so I added a minWidth property to resolve this.

stevenwang213 (approved):

PR #421, opened 2022-05-28
Ron intangible admin form fix

ronBP95:
Resolved issue with "Tangible Form pops up as default when logged into Admin Account even if the 'Add Intangible Time Entry Form' is clicked.

Line 58 was commented out and provides a fix for the solution however I did find that there is an alternative solution. We can remove 'dataIsTangibleTimelog' as a permission for the Administrator role in utils/permissions. 

Tested:
- Line 58 in timelog.jsx was commented out
- Add intangible form entry now works when logged into Admin account
- Tangible form entry works when stopping timer
- Intangible and Tangible forms are displaying/trigger correctly and was tested on a 'Volunteer' account.

### DIFF
--- a/src/components/Timelog/Timelog.jsx
+++ b/src/components/Timelog/Timelog.jsx
@@ -55,7 +55,7 @@ class Timelog extends Component {
     this.openInfo = this.openInfo.bind(this);
     this.data = {
       disabled: !hasPermission(this.props.auth.user.role, 'disabledDataTimelog') ? false : true,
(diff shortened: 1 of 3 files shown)


stevenwang213 (approved):
The following tests were tested:
1. login as admin and click 'add intangible time entry', the popup form displays 'intangible' as expected.
2. login as admin and add time entry from the timer, the popup form displays 'tangible' as expected.
3. login as volunteer and do test 1&2, both show correct info on the popup form.

PR #422, opened 2022-05-30
change active inactive button behavior in user profile page

stevenwang213:
This PR resolves the bug that the previous paused user will automatically come back after being inactive from their profile page.
In this PR, I copy and reuse the logic of "active inactive button" on user management page to user profile page.

To test this PR, you have to inactive a user that had been paused and resumed previously e.g. "Jerry Volunteer1" from their profile page. And make sure this user does not become "active" after being deactivated.

image
image


migfernandes01 (approved):
Looks good to me!

migfernandes01 (approved):
Everything looks good!

PR #425, opened 2022-05-31
Reverted change from font applied to body of application

ronBP95:
PR issued to revert CSS changes made to PeopleReport.css that affect application font. This change will bring the original font of the application back.


veratim1996 (approved):
Now I approve, but ideally there shouldn't be common styles in this file.

PR #428, opened 2022-06-01
Change three buttons 'update password', 'cancel', 'save change' display logic

stevenwang213:
To the reviewer:

This PR changes the display logic of the three buttons 'update password', 'cancel', and 'save change' on the User Profile page.
When non-Admin users view their own profile, these three buttons should only exist when they are on the "Basic Information" tab.
And when admin users view everyone's profile,  these three buttons should exist on every tab.

### DIFF
--- a/src/components/UserProfile/UserProfile.jsx
+++ b/src/components/UserProfile/UserProfile.jsx
@@ -331,15 +331,13 @@ const UserProfile = props => {
   // const isUserAdmin = requestorRole === 'Administrator';
   // const canEdit = hasPermission(requestorRole, 'editUserProfile') || isUserSelf;
   let canEdit;
-  if(userProfile.role !== 'Owner'){
+  if (userProfile.role !== 'Owner') {
     canEdit = hasPermission(requestorRole, 'editUserProfile') || isUserSelf;
   } else {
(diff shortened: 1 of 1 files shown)


migfernandes01:
@stevenwang213 It looks great, good job, the only problem I see is the following:
Whenever I use an Owner and see other profiles I can't see the save/cancel buttons

Owner seeing other profile with dev:
owner seeing other profile dev

Owner seeing other profile with this branch:
owner seeing other profile local

I'm not sure if you saw it or not, but I believe roles like Owner should be able to change and save changes on any other profile.
I only checked on Owner, but you might want to check every other role (Admin, Manager, Mentor, and Core Team).
Let me know how it goes, so I can get it approved soon!

stevenwang213 (commented):
@migfernandes01 Thanks for pointing out the problem. I have fixed it in the new commit, the owner should be able to see the save/cancel button on everyone's profile now. I also test every other role(Admin, Manager, Mentor, and Core Team) and they all work well. Let me know if you find any bugs :)

one-community:
Tested this all on the Dev and it looks good and works well. When on the Volunteering Times tab for a Volunteer though, all the fields look editable. Can we make those NOT look editable for all classes except Owner and Admin?

PR #429, opened 2022-06-01
Timer sync in all Buttons

ivany9:
Currently a person can start a timer, pause it, then open a new tab with a timer and it will show the same time on the 2nd one (this is correct). If they start the second timer and leave the first paused though, they can then click “stop” and log their time on the paused one and keep the 2nd one going. To eliminate this, the timer needs to sync all timers when the “Submit” button is clicked. 

request


thienphuvn95:
PR looks good to me. This is how I test:
- Open two hgn, click start timer on first page, the second page also started by itself (this is correct)
image
- Click pauses on 2nd page or 1st page both stopped all hgn
image
- Stop timer on 2nd page or 1st page both reset timer, added tangible time
image
-Log out on 2nd page. The 1st page timer still running
image
-Log in on 2nd page, the timer is delayed a bit, but quick change to 1st page timer
image
- Log in 2nd page with different user, the timer on both pages are different
image

ivany9:
@thienphuvn95   The Timer is Synchronized but the delay is due to the time response.  It needs to fetch a lot of things such as a userid,  the time, etc.  At the moment I'm working to pause the timer when the other tab is closed for the same user.

thienphuvn95:
- I started timer, logged out of hgn, and backed to hgn for 10 minutes but the timer was still running. (it should reset by limit like 5 minutes)
image

one-community:
Phu is correct: The app should stop the timer if a person’s computer goes to sleep, is put to sleep, or logs off
It's not checking if someone is working, it's checking if their computer goes to sleep or logs off or closes the window. In short, the app needs to be open or the app should shut off after 5 minutes. This is also to help people who may forget to stop the app.

ivany9:
Working in this Bug.
 **Timer Auto-stop and Logout Functionality**
The app should stop the timer if a person’s computer goes to sleep, is put to sleep, or logs off
It's not checking if someone is working, it's checking if their computer goes to sleep or logs off or closes the window. In short, the app needs to be open or the app should shut off after 5 minutes. This is also to help people who may forget to stop the app.

one-community:
Following up on this. I’m getting reports from people that the current timer isn’t working right, it keeps resetting even when someone is working consistently. So I’d like to get this checked in as soon as possible.

one-community:
THIS PR IS NOT MERGEABLE - PLEASE LEAVE IT UNTIL I CAN FIND SOMEONE ELSE TO TAKE OVER ON THESE FIXES.

PR #432, opened 2022-06-02
impl single task wbs page

EvianTan:
Implement clicking the task on dashboard, will lead to the pettier WBS page with only the task's details.
 
The below image is after clicking the "HGN Software Development" task from management-dashboard.
image

**Note:**
**1. The left-up back button function is still need to be discussed
2. Now this PR only works for level-1 tasks on management-dashboard**


migfernandes01 (approved):
Looks good to me!

one-community:
Looks good here. Can't test it though because my assigned tasks are not showing up anymore on my Dashboard. So can't click them to see where they go. Not sure if they have disappeared due to this PR or some other one because they aren't something we're currently using, so just now noticed this.

PR #435, opened 2022-06-04
We have to use auth.user.role instead userProfile.role because some r…

veratim1996:
…oles can edit other people's profiles. userProfile.role is about currently opened profile, not the active user.

### DIFF
--- a/src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
+++ b/src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
@@ -28,6 +28,7 @@ import ReminderModal from './ReminderModal';
 import axios from 'axios';
 import { ApiEndpoint } from '../../../utils/URL';
 import hasPermission from 'utils/permissions';
+import { getTimeEntryFormData } from './selectors';
 
 /**
  * Modal used to submit and edit tangible and intangible time entries.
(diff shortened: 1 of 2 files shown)


one-community:
Hi Vera, 

So others can check your PRs more easily, please copy/paste the original problem into the description. As this is written here, I'm not clear what this PR is actually fixing. I think it is the one we've been talking about, but I'm not sure and someone not part of our conversations would be even less sure. So please include what the problem was AND how this PR fixes it.

migfernandes01:
Found an interesting behavior. At first I was clicking the circle close to a user profile on the leaderboard component and going to a specific user dashboard. From there I was able to edit the dates for the logged times. But when I clicked the progress bar and going to a user's timelog I got this error(I tried to edit the date using the Admin and Owner roles):
error saving timelog after editing date

Let me know if this PR only fixes the edition for the dashboard component or if I'm doing anything wrong (I already pulled the latest version of the backend and frontend using your branch)

one-community:
No, it should allow editing of the dates and tangible/intangible status from both places.

ivany9:
Working in this Bug at the moment:
**Timer Auto-stop and Logout Functionality**
"The app should stop the timer if a person’s computer goes to sleep, is put to sleep, or logs off
It's not checking if someone is working, it's checking if their computer goes to sleep or logs off or closes the window. In short, the app needs to be open or the app should shut off after 5 minutes. This is also to help people who may forget to stop the app."

stevenwang213 (approved):
I can edit and add time logs from both places. Looks good to me!

PR #442, opened 2022-06-08
add scrollbar and change display order of blue squares

stevenwang213:
In this PR, I added a scrollbar to the blue square area and moved the "add blue square" button to the first place. Also, I reversed the order of the blue square. Now the newest one will show first. So, the admin can directly find/edit the newest blue square without scrolling to the very end.
image


thienphuvn95:
image
PR looks good to me

thienphuvn95 (approved):

PR #443, opened 2022-06-08
make some tab non-editable to non-admin user

stevenwang213:
In this PR, I made all tabs non-editable to the non-Admin users except the basic information tab in user profile.
This is an admin view of other's profile. All tabs are editable.
Capture

This is non-admin view of other's profile. All tabs are non-editable except Basic Information tab.
image


thienphuvn95:
image
Admin
image
User.
Approved

thienphuvn95 (approved):

PR #447, opened 2022-06-12
Irenedev

thienphuvn95:
Fix Admin, User permission to change, fix showing categories hours.
Also, this PR will have new bug as Admin could change tangible hours but those change needs to be saved in database and reload to frontend. Will keep working on this.


ronBP95:
Looking at this

ronBP95:
I tested out several categories and was unable to get the time to add to the statistics in:
- Food
- Stewardship
- Education

I added a few tangible time entries to see if the values would update in the categories but they were not showing up as intended. Stewardship should have at least 5 hours of tangible time entries. 

phu pr check

ronBP95:
Here are the entries for reference:

phu check

EvianTan:
I understand this PR just working on "changing the different category hours would effect the total one" on Volunteering Times Tab from people's profile page. This PR works great on changing the total hours concurrently when category hours changed. Thank you Phu!

About the functionality, I found something that might can be improved:
1. Category hours cannot be negative
I changed the housing hours to be negative and saved, it saved successfully, but the negative hours didn't contribute to the total tangible hours
image

2. The logic of updating the hours from keyboard inputs is weird
Here we have two ways to change the hours, use the two triangle buttons to adjust hours by +1/-1 or input by keyboard.
The triangle buttons work well, but the inputs by keyboard is not that convenient. The way it read inputs from keyboard is very weird and seems it also cannot read decimal point, so you may need to update the hours by integer part and decimal part separately.
My suggestion is we can get rid of those two triangles, only use inputs from keyboard, and allow user to input number with decimals.

3. Hovering info
When my mouse hovering on the little triangle buttons, some how it show this info. If it's unrelated, we can just say something like "please give valid input". 
image

If you have any questions/concerns on above, feel free to tell me. I'd love to discuss them 😄

src/components/Reports/PeopleReport/index.js
```diff
@@ -1 +1 @@
-export { default as PeopleReport } from "./PeopleReport";
+export { default as PeopleReport } from "./PeopleReport/PeopleRe";
```
EvianTan, line 1: Looks like a typo to me, consider revert it.

src/components/UserProfile/VolunteeringTimeTab/VolunteeringTimeTab.jsx
```diff
@@ -19,6 +21,9 @@ const StartDate = (props) => {
       onChange={(e) => {
         props.setChanged(true);
         props.setUserProfile({ ...props.userProfile, createdDate: e.target.value });
+        <Alert color="warning">
```
EvianTan, line 24: I think we can do a bit refactoring here. I removed all other `<Alert />` elements just kept the last one, it also worked.

src/components/UserProfile/VolunteeringTimeTab/VolunteeringTimeTab.jsx
```diff
@@ -5,6 +5,8 @@ import { capitalize } from 'lodash';
 import style from '../UserProfileEdit/ToggleSwitch/ToggleSwitch.module.scss';
 import { ENDPOINTS } from 'utils/URL';
 import axios from 'axios';
+import Alert from 'reactstrap/lib/Alert';
```
EvianTan, line 8: I think we already have this alert part in `UserProfile.js`. I commented out all the <Alert /> things in this file, it did show the alert "Please click on "Save changes" to save the changes you have made."...
image
I also found your screenshot didn't have that alert. Please go and check other tabs(Teams, Projects,...), if both of them don't have that alert on your dark mode, that would be another bug.
image

image

src/components/UserProfile/VolunteeringTimeTab/VolunteeringTimeTab.jsx
```diff
@@ -94,27 +105,37 @@ const WeeklyCommitedHours = (props) => {
...
     />
   );
 };
-
 const TotalTangibleHours = (props) => {
+  var sum=0;
+  props.userProfile.categoryTangibleHrs.forEach(object=>{
+    sum+=Number(object.hrs);
+  }) 
+  console.log(props.userProfile);
```
EvianTan, line 122: Please remove this debug output.

src/components/UserProfile/VolunteeringTimeTab/VolunteeringTimeTab.jsx
```diff
@@ -232,50 +253,58 @@ const ViewTab = (props) => {
...
+            // canEdit={canEdit}
             role={role}
           />
         </Col>
       </Row>
 
-      {props?.userProfile?.hoursByCategory
-        ? Object.keys(props.userProfile.hoursByCategory).map((key) => (
-            <React.Fragment key={'hours-by-category-' + key}>
+      {props?.userProfile?.categoryTangibleHrs.map((key) => (
```
EvianTan, line 262: Nice updates on this part 😄

EvianTan:
Nice update on the functionality!!

**Verified changes:**
1. Prevent inputting negative number for times ✅ 
2. Keyboard input for changing times works well ✅ 
3. Merged a lot of changes from development branch ✅ 👍 

**Things need to fixed before this PR close:**
1. Resolve this run test error: ❗ 
image
This one you can just revert all your changes in `src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx`.
Actually I made the changes using `projectOrTaskOptions`, because we were working on this feature to make the site can log time to projects and also tasks. If you have any questions on that, just let me know.

2. Some unrelated changes need reverted: ❗ 
`src/components/Reports/PeopleReport/index.js`
`src/components/Timelog/TimeEntry.jsx`
Usually we need keep minimum of our changes 😄 

**Things can improve, consider filing into a bug:**
1. Take two decimals for times instead of 13 decimals..
image

thienphuvn95:
2. The reason why I kept 13 decimal (not really 13, it depends on database like any default number) because if I define those field to be 2 decimals, it will not be editable with correct number

EvianTan:
> 2. The reason why I kept 13 decimal (not really 13, it depends on database like any default number) because if I define those field to be 2 decimals, it will not be editable with correct number. If anyone know how to fix this, I'm really happy to get their help!

I fully understand this one. Just put it in the bug list 😄

PR #453, opened 2022-06-14
add tasks as options when log time

EvianTan:
Create the ability to log time for tasks (currently we can only log time for projects)
**This PR is the first part of changes for above feature, only UI changes. Backend/database changes and related "Time Log listing table" changes will come into next PR**.

The below images show log tangible/intangible time the popup table, they do have tasks assigned to current user as dropdown options
image
image

### DIFF
--- a/src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
+++ b/src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
@@ -26,7 +26,7 @@ import AboutModal from './AboutModal';
 import TangibleInfoModal from './TangibleInfoModal';
 import ReminderModal from './ReminderModal';
 import axios from 'axios';
-import { ApiEndpoint } from '../../../utils/URL';
+import { ENDPOINTS } from '../../../utils/URL';
 import hasPermission from 'utils/permissions';
 import { getTimeEntryFormData } from './selectors';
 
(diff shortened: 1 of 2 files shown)


src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
```diff
@@ -108,6 +109,15 @@ const TimeEntryForm = (props) => {
       .catch((err) => {});
   }, []);
 
+  useEffect(() => {
+    axios
+      .get(`${ApiEndpoint}/tasks/userProfile?members=${userId}`)
```
esk47, line 114: You should define an endpoint for this URL (I think in URL.js) so that it is reusable. If one already exists, you should use that.

src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
```diff
@@ -108,6 +109,15 @@ const TimeEntryForm = (props) => {
...
+      .then((res) => {
+        setTasks(res?.data || []); 
+      })
+      .catch((err) => {});
```
esk47, line 118: A warning popup here could be good, but it is not necessary.

esk47 (changes requested):
Everything looks good and route is worth fixing.

src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
EvianTan: **Addressed.**

src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
EvianTan: **Addressed.**

esk47 (approved):
Looks good

PR #454, opened 2022-06-15
implement progress bar for tasks

kimsj23:
This PR is to create a progress bar to track hours worked and estimated hours on a per task basis assigned to an individual.

Below is the feature requirement:
image
Example of end product:
image

Below is the current state:
image

Note: Since the task hours logged functionality is not currently in place, the progress bars will be 0 / empty for all tasks.

To test functionality, you can use the code in green below by uncommenting out and commenting out the code already in place in this PR. To test specific colors, replace code in **color** attribute with a number (using the above color listing requirement as reference).
image

### DIFF
--- a/src/components/TeamMemberTasks/TeamMemberTasks.jsx
+++ b/src/components/TeamMemberTasks/TeamMemberTasks.jsx
@@ -351,7 +351,7 @@ const TeamMemberTasks = props => {
                   
                   {/* progress bar implementation */}
                   <td className='team-task-hours'>  
-                    <div key={`${task._id}${index}`}>
+                    <div>
                       <span>
                       {`${parseFloat(task.hoursLogged.toFixed(2))} 
                         of 
--- a/src/components/TeamMemberTasks/TeamMemberTasks.jsx
+++ b/src/components/TeamMemberTasks/TeamMemberTasks.jsx
@@ -5,7 +5,7 @@
 import React, { useState, useEffect } from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
-import { Table, Button, Modal, ModalHeader, ModalBody, ModalFooter } from 'reactstrap';
+import { Table, Button, Modal, ModalHeader, ModalBody, ModalFooter, Progress } from 'reactstrap';
 import moment from 'moment';
 import _ from 'lodash';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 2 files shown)


src/components/TeamMemberTasks/TeamMemberTasks.jsx
```diff
@@ -18,6 +18,8 @@ import { fetchAllManagingTeams } from '../../actions/team';
 import { getUserProfile } from '../../actions/userProfile';
 import Loading from '../common/Loading';
 import DiffedText from './DiffedText';
+import { getcolor } from '../../utils/effortColors'
```
EvianTan, line 21: Nice reuse of the color part!

EvianTan:
Nice logic here!
I did find couple questions need to be discussed and fixed.

1. The alignment of task and its progress number and progress bar.
See the the below image, "160 of 1963.33" takes two lines, it can lead all below progress number and bar mismatch. Consider fix this problem.
image

2. The progress bar color
From the description and example of end product image above, logged 4 hours for one task will give the progress red color, right?
image
The only thing I think still need to be discussed is the situation logged time is larger than estimated hours. Does it need any special color for progress bar?

src/components/TeamMemberTasks/TeamMemberTasks.jsx
kimsj23: Thanks for the review.

1. Are you looking at this on mobile view?
2. No, the color is based on the hours remaining on a task, not the amount of hours logged into a task based on the requirement above. I am not sure the end product image is correct here.
3. If the logged time is greater than the estimated, it should remain a red color based on requirement above.

EvianTan: 1. I looked this on my laptop. When I resized the window it looks like this. 
2. Yep I also saw this point and got focused, **maybe check with Jae.**
3. Understood! Great!

one-community: The text in your first image is correct. The second posted image does not match the text. Colors for those bars should match how much time _remains_ for the task.

kimsj23: The problem for #1 looks to be fixed now and new commit made. Please review again. 

Thanks

EvianTan:
Pretty good changes! Thank you!
image

EvianTan (approved):

PR #455, opened 2022-06-15
Reconfigured css properties to fix the css slider bug

ronBP95:
For reference:
bug

### DIFF
--- a/src/components/UserProfile/Badge.css
+++ b/src/components/UserProfile/Badge.css
@@ -89,7 +89,7 @@
 .badge_featured_container {
   display: flex;
   flex-wrap: nowrap;
-  overflow-y: none;
+  overflow-y: hidden;
 }
 
 .badge_md_image_container {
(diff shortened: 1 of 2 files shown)


ronBP95:
Note:

I've added an id to one of the bootstrap card elements in hopes of overwriting a min-width: 0 (now min-width: 475px) which I found to be the reason for the bug potentially. 

For reviewer:

If the issue still persists, please inspect the badge container element and provide a snippet of the CSS that is applied. I was experiencing a similar but different display issue than the image that was provided above.

ronBP95:
@veratim1996 any progress on this? I know you mentioned you were having trouble replicating it on your browser but I'm wondering if you found anything different on your end

veratim1996:
> 

I asked you about that in Slack. I'm waiting for answer:)

veratim1996 (approved):

PR #458, opened 2022-06-18
update opened week time entries after editing a time entry

veratim1996:
The PR fixes the bug that appears when you edit a time entry and change the date so the time entry must move to another week. 
The wrong behavior was that the edited time entry was moved to another week, but did not disappear from the edited week's entry list until the page was refreshed. Now it works without refreshing.

### DIFF
--- a/src/actions/timeEntries.js
+++ b/src/actions/timeEntries.js
@@ -67,12 +67,12 @@ export const postTimeEntry = (timeEntry) => {
   };
 };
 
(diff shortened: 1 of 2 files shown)


ronBP95:
Taking a look at this

ronBP95:
Tried making a time entry in a previous week and changing it to the present week and it seamlessly transitioned without having to refresh the page. Nicely done
vera check
vera check 2

!

ronBP95 (approved):

PR #461, opened 2022-06-23
filter time entries by task

EvianTan:
This PR is the frontend changes for implementing "filter time entries by task"

Detail changes:
1. make the task name of each related time entries present
~~2. per Jae's suggestion, change the filtering function to only filter by task~~
3. change related test based on the failed check

Pictures:
image

### DIFF
--- a/src/components/Timelog/TimeEntry.jsx
+++ b/src/components/Timelog/TimeEntry.jsx
@@ -42,6 +42,7 @@ const TimeEntry = ({ data, displayYear, userProfile }) => {
 
   useEffect(() => {
     axios
+      // Note: Here taskId is stored in projectId since no taskId field in timeEntry schema
       .get(ENDPOINTS.GET_TASK(data.projectId))
       .then((res) => {
         setTaskName(res?.data?.taskName || "")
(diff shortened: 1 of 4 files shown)


src/components/Timelog/TimeEntry.jsx
```diff
@@ -25,6 +31,24 @@ const TimeEntry = ({ data, displayYear, userProfile }) => {
...
+      .get(ENDPOINTS.PROJECT_BY_ID(data.projectId))
+      .then((res) => {
+        setProjectName(res?.data?.projectName || "")
+      })
+      .catch(err => console.log(err));
+  }, [])
+
+  useEffect(() => {
+    axios
+      .get(ENDPOINTS.GET_TASK(data.projectId))
```
kimsj23, line 45: Should this be `data.taskId` instead?

kimsj23 (changes requested):
Good stuff. Looks to be working well. I did notice one thing that I commented on below. Can you take a look at it?
Thanks

kimsj23 (approved):

PR #475, opened 2022-07-07
Modal for checking User Dupe was completed but some logic is required!

ronBP95:
The modal was already built out. Just need to connect it to the userProfile object and run a check.

ex. if (userProfile.first[i] === input) {
  triggerModal();  <---- function to triggerModal and its own flow before posting data
}

Screenshot 2022-06-25 231749


ronBP95:
Not sure why there are additional commits. Closing this and reopening a new PR

PR #476, opened 2022-07-07
 Nested Modal mostly completed. Logic needed for triggering

ronBP95:
The developer just needs to hit the userProfile object as such:

ex. if (userProfile.first[i] === input) {
triggerModal(); <---- function to triggerModal and its own flow before posting data
}

Screenshot 2022-06-25 231749


thienphuvn95:
Hi Ron,
I checked your change on user management. Does it only include the trigger modal or matching first name/last name before those modal appear?

thienphuvn95 (approved):

PR #477, opened 2022-07-07
New Fix for intangible admin form

ronBP95:
The intangible time entry form will now work for admin accounts with role === 'adminLinks'. This change now defaults to the intangible time entry form when clicking 'Add Intangible Time Entry" but comes with a small visual bug that shows a 'checked' tangible box when the form is really for intangible time entries.

Intangible Time Entry for Admin


ronBP95:
Some logic needs to be played with in the 'checked:' property on line 537 in the TimeEntryForm.jsx component. I tried several changes to make it work but it ended up breaking the button instead. This it the closest that I can bring the component to it's intended function.

ronBP95:
These changes could likely be merged without affecting volunteering accounts. Only accounts with role === 'adminLinks' (which are the administrative accounts should be affected by this PR.

thienphuvn95:
The Design is correct but the logic seems odd to me.
When you click "Add intangible time entry", the modal should appear with Intagible form and unchecked tangible box.
if you check the box, it should switch to tangible.

one-community:
Yes, that's correct Phu. Ron has wrapped up his work with us, can you make the needed changes?

thienphuvn95:
Yes, I can

one-community:
Thank you. If needed, the same will go for Ron's other PR.

thienphuvn95 (approved):

PR #478, opened 2022-07-08
Fix modal displaying task diff for other users

esk47:
Bug fix for task notification modal displaying task notifications for other users.

To test:
1. Add/Edit a task and assign it to 2 or more users
2. Click on the red bell that appears. Only 1 task notification should be there.


kimsj23 (approved):
Looks good to me. Approved!

PR #479, opened 2022-07-08
border-radius was cutting off the edges

veratim1996:
The problem: image

### DIFF
--- a/src/components/UserProfile/Badge.css
+++ b/src/components/UserProfile/Badge.css
@@ -12,7 +12,6 @@
 }
 
 .badge_image_md > img {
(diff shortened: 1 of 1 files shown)


kseniaxu0704:
Looks good to me!

kseniaxu0704 (commented):
Looks good to me. Here just one question: so by removing the border-radius, we can let the icon fit? Or should we re-set this border-radius to another percentage?

veratim1996:
I see no reason to keep border-radius

kseniaxu0704 (approved):

PR #486, opened 2022-08-09
Yiyun create single task component

EvianTan:
This PR is implementing "click a task from Management-Dashboard would jump to its detail page".

This functionality was implemented for only level 1 tasks before by filtering the task from whole WBS tasks page. This PR is creating a new component called "SingleTask", which contains all needed detail info of a task. 
Take the "2.4 Finish Tree House Village Book Edits" assigned to "Jae Test" for example. See below images, the first one is before, second one is after.
image
image

**Main changes of this PR:**
1. create "SingleTask" component and use it
2. revert the temporary implementation of showing task in `src/components/Projects/WBS/WBSDetail/WBSTasks.jsx`

**Note this PR is only for the jumping and showing details of task, below features are not included and will come in next PRs:**
1. Click the top left ◀️ button will jump a page contains all tasks in the same folder of current one
3. Admin/Owner users can directly edit this task or add a new task on "Single Task page" (still need to be discussed before implementing)

### DIFF
--- a/src/components/Projects/WBS/SingleTask/SingleTask.js
+++ b/src/components/Projects/WBS/SingleTask/SingleTask.js
@@ -1,11 +1,119 @@
+import React from "react";
+import { useState, useEffect } from "react";
+import ReactTooltip from "react-tooltip";
+import axios from 'axios';
+import { NavItem, Button } from "reactstrap";
+import { Link } from "react-router-dom";
+import { ENDPOINTS } from 'utils/URL';
 
 const SingleTask = (props) => {
   console.log("single task props is: ", props)
   const taskId = props.match.params.taskId;
   console.log("task id is: ", taskId)
 
-  return (
-    <p>Hello world</p>
+  const [task, setTask] = useState({});
+
+  useEffect(() => {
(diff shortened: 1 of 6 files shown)


src/components/Projects/WBS/SingleTask/SingleTask.js
```diff
@@ -0,0 +1,204 @@
...
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <th scope="row">{task.num}</th>
+              <td>{task.taskName}</td>
+              <td>{task.priority}</td>
+              <td className="desktop-view">
+                {task.resources
```
andrewpaulino, line 97:
```suggestion
                {task?.resources &&
```
Also make this optionally chained, since you are testing for the object property.

You can use the `&&` and not need to return null in this case.

EvianTan: Thanks for that review! I'll do that!

esk47 (approved):
Approved.

Everything works well.

PR #488, opened 2022-08-11
Fix tangible checkbox

veratim1996:
When Intangible Time is logged by Admins/Owners, the “Tangible time” box was checked but should be unchecked by default. I deleted the `dataIsTangibleTimelog` permission because it isn't needed in any scenario.

Screenshot_1

### DIFF
--- a/src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
+++ b/src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
@@ -80,7 +80,7 @@ const TimeEntryForm = (props) => {
   const { userProfile, currentUserRole } = useSelector(getTimeEntryFormData);
 
   const dispatch = useDispatch();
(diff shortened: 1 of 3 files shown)


andrewpaulino (commented):
Why don't we need the permission anymore just curious?

veratim1996:
> Why don't we need the permission anymore just curious?

I think, we never did. This permission was true only for admins and owners. And it only affected whether time would be tangible or not. But tangible/intangible time isn't related to the user's role, is it? The only thing that depends on the user's role is whether the user can toggle the "tangible checkbox", and for that we have other permission.

EvianTan:
Verified all the changes. Tested logging as all different roles, log tangible/intangible time entries. They worked well and behaviors were as expected. Good Job!!

EvianTan (approved):

PR #498, opened 2022-09-02
Remove ! from Weekly Committed Hours

Programmingrelated1999:
This PR is for removing ! from Input field for Weekly Committed Hours which is a bug fix. 

There were two changes made in WeeklyCommittedHours and TotalTangibleHours. 

* The "!" showing in VolunteeringTimesTab.jsx was due to invalid field in <Input> set to !props.isUserAdmin() which was not passed to the component, and thus is always undefined. Therefore ! of undefined value makes it always true, and thus shows a warning symbol in the input field. 
* In TotalTangibleHours component, there is an if statement with condition set to props.isUserAdmin(). The if statement is removed since it is always true. The previous code is commented out.

The input field shows no warning as shown below: 
Administrator_weekly_committed_hours_bug

### DIFF
--- a/src/components/UserProfile/UserProfileEdit/UserProfileEdit.jsx
+++ b/src/components/UserProfile/UserProfileEdit/UserProfileEdit.jsx
@@ -84,7 +84,7 @@ class UserProfileEdit extends Component {
     // console.log('edit profile, state:', this.state);
   }
 
-  toggleTab = (tab) => {
+  toggleTab = tab => {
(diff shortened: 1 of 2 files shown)


kseniaxu0704:
Hi Kaung, code seems great, but I want to test it on my local host. I am trying to fetch and switch your branch, but it seems like the "!" in your branch name blocks me from doing so. Could you please change your branch name? leveraging "-" or "/" perhaps would be a better choice to separate each word. Thx

PR #500, opened 2022-09-04
Fix to remove HTML tags and &nsbp; being displayed in red bell diff

esk47:
Copy and pasting text would sometimes include HTML tags and nbsp in the text being displayed when clicking on the red bell to open the modal.


EvianTan:
Verified changes:
Before:
image
After:
image

EvianTan (approved):

PR #501, opened 2022-09-04
Timezone API UI fix-updating timezone state when using GetTimeZone button, minor linter changes

rutvij2811:
Issue: Updating the profile with the new timezone is not reflected in the UI or the database.

Root cause: Not updating the state value of TimeZone, when the "Get Time Zone" button is clicked. Hence it stored only the default value at the backend.

Fix: Whenever the button "Get Time Zone" is clicked, it should change the state for "timezone" also along with the "timezoneFilter" (Part of the dropdown component).

saveTImeZoneFixUI

### DIFF
--- a/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
+++ b/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
@@ -176,7 +176,7 @@ class AddUserProfile extends Component {
                       <PhoneInput
                         country={'us'}
                         value={phoneNumber}
-                        onChange={(phone) => this.phoneChange(phone)}
+                        onChange={phone => this.phoneChange(phone)}
                       />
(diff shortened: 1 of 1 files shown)


Programmingrelated1999:
It looks good.

Programmingrelated1999:
My Bad. I could not get the result from the snip video on my computer. I am not sure if this is the problem with the API. 
PR_review

rutvij2811:
You will have to add the timezone api key in the local env first before it works.

Programmingrelated1999:
Got it working! Everything looks good. I will go ahead and approve it.

Programmingrelated1999 (approved):
Everything looks good. I was able to update the time zone based on location.

one-community:
Get Timezone button is no longer working for me on the Dev after merging this PR.

PR #504, opened 2022-09-07
add edit button for singleTask page

EvianTan:
This PR is the **first half** of adding "Edit" button on singleTask page.
**This feature only works for level 1 tasks now.**

Main changes explained:
1. `src/components/Projects/WBS/SingleTask/SingleTask.js`
add column in the singleTask table, and put the `EditTaskModal` there
2. `src/components/Projects/WBS/WBSDetail/EditTask/EditTaskModal.jsx`
- add optional chaining operators (?.) to remove the error when fetching task data the task object is empty
- trim code 

add the edit button for single task page

### DIFF
--- a/src/components/Projects/WBS/SingleTask/SingleTask.js
+++ b/src/components/Projects/WBS/SingleTask/SingleTask.js
@@ -9,6 +9,7 @@ import { Editor } from '@tinymce/tinymce-react';
 import { Link } from "react-router-dom";
 import { ENDPOINTS } from 'utils/URL';
 import { getUserProfile } from "actions/userProfile";
+import EditTaskModal from "../WBSDetail/EditTask/EditTaskModal";
 import hasPermission from "utils/permissions";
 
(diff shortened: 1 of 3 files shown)


esk47 (dismissed):
As you said, it only works with level 1 tasks.

Approved but write down the bug in the bugs doc and/or management dashboard doc

EvianTan:
For the second commit, I implemented this "Edit" button also work for level 2/3/4 tasks. 

Main changes:
1. src/components/Projects/WBS/WBSDetail/EditTask/EditTaskModal.jsx
get the task props by `GET_TASK_BY_ID` function, instead of original filtering things, so that we can get the level 2/3/4 tasks
2. src/components/Projects/WBS/WBSDetail/Task/Task.jsx
Add a ternary operator to avoid the empty props error when getting data into `Task.jsx`

See below image:
image

esk47 (approved):
Bug seems fixed

PR #505, opened 2022-09-08
Fixed the issue, now working on both UserCreation and while Editing

rutvij2811:
I have fixed the issue where the "Get Time Zone" button didn't work while creating a new user. The error was caused while changing the states. (I have not worked with class-based state updates in the past. Basically, it threw a Reference Error).

Also, there was a small typo, "timezone" was used instead of "timeZone" due to which the timeZone payload was not sent to the database while creating a new user.

(Note: I will put the screenshots later if necessary)


Programmingrelated1999:
I was able to get it working on my computer when creating a new user. I will go ahead and approve this pull request.

Programmingrelated1999 (approved):
It works for user creation.

PR #514, opened 2022-09-15
Fix TimeZone problem for finalday and add a 5s notfication banner aft…

kseniaxu0704:
There are two goals for this PR:
1) Fix the bug for being unable to set a final day when the final day is tomorrow 
2) Create a 5 seconds notification banner when setting or deleting the final day:

Fix bug for unable to set a final day when the last day is tomorrow 

1) In the past, it could not set a final day tomorrow due to the different time zone. Therefore i use moment().isBefore(moment(finalDayDate) to keep time in the same time zone;

Another change for this PR is to give a notification banner when setting or deleting the final day:

It used to direct reload the page when clicking the saves button or delete button, but now I write a function with "toast" toast.success("This user's final day has been deleted.")/toast.success("This user's final day has been set.") to offer a 5 seconds notification - 
Screen Shot 2022-09-14 at 10 09 25 PM

### DIFF
--- a/src/components/UserManagement/SetUpFinalDayButton.jsx
+++ b/src/components/UserManagement/SetUpFinalDayButton.jsx
@@ -25,8 +25,10 @@ const SetUpFinalDayButton = (props) => {
     if (isSet) {
       updateUserFinalDayStatus(props.userProfile,"Active",undefined)(dispatch)
       setIsSet(!isSet);
-      toast.success("This user's final day has been set.");
-      window.location.reload();      
+      toast.success("This user's final day has been delete.");
+      setTimeout(function(){
+        window.location.reload();
+     }, 5000);
     } else {
       setFinalDayDateOpen(true);
     } 
@@ -40,8 +42,10 @@ const SetUpFinalDayButton = (props) => {
     updateUserFinalDayStatus(props.userProfile, "Active", finalDayDate)(dispatch);
     setIsSet(true);
     setFinalDayDateOpen(false);
-    toast.success("This user's final day has been deleted.");
-    window.location.reload();
+    toast.success("This user's final day has been set.");
+    setTimeout(function(){
+      window.location.reload();
+   }, 5000);
   };
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/UserManagement/SetUpFinalDayButton.jsx
```diff
@@ -25,8 +25,10 @@ const SetUpFinalDayButton = (props) => {
     if (isSet) {
       updateUserFinalDayStatus(props.userProfile,"Active",undefined)(dispatch)
       setIsSet(!isSet);
-      toast.success("This user's final day has been set.");
-      window.location.reload();      
+      toast.success("This user's final day has been delete.");
```
brucelin006, line 28: deleted?

src/components/UserProfile/BasicInformationTab/BasicInformationTab.jsx
```diff
@@ -566,7 +566,7 @@ const BasicInformationTab = (props) => {
           <Label>
             {
             userProfile.endDate
-            ? 'End Date ' + moment(userProfile.endDate).format('YYYY-MM-DD')
+            ? 'End Date ' + userProfile.endDate.toLocaleString().split('T')[0]
```
brucelin006, line 569: It seems that the format isn't consistent with line 557 where it writes
`moment(userProfile.reactivationDate).format('YYYY-MM-DD')`, whereas the original code is consistent. Would this be a potential problem in the future? Or is it better to also change line 557 to `userProfile.reactivationDate.toLocaleString().split('T')[0]`?

brucelin006 (commented):
Perhaps I am wrong, but please check out the comment on the `src/components/UserProfile/BasicInformationTab/**BasicInformationTab.jsx**` file?

src/components/UserManagement/SetUpFinalDayButton.jsx
kseniaxu0704: Sure, will revise here. Is there anything else that I need to revise?

src/components/UserProfile/BasicInformationTab/BasicInformationTab.jsx
kseniaxu0704: Thanks, good catch, but I want to remain it here, due to fixing the bug of time zone issue. As for the reactivationDate, we can add another task on aligning it, not for this PR, which is beyond the scope.

brucelin006 (approved):

Programmingrelated1999:
Everything looks great.

PR #516, opened 2022-09-16
Pie charts on the people report page

veratim1996:
PR with two pie charts: one displays info about projects with committed hours, second displays info about tasks with committed hours. The tasks pie chart has complex display conditions in order to avoid displaying too long list of the tasks. If there is a lot of tasks, the pie chart shows just the last modified tasks and the 'View all" button. Clicking on the button opens a modal with a pie chart that displays information about all the tasks, not just few of them. 

Here are the conditions for displaying the tasks pie chart:
1. If the number of tasks is less than or equal to the number of projects, all tasks are displayed, the "View all" button is hidden
2. If the number of tasks is more than the number of projects but less than or equal to 4, all tasks are displayed, the "View all" button is hidden
3. If the number of tasks is more than the number of projects an more than 4, then only 4 last modified tasks are displayed, the "View all" button is available.

If there are less than 8 tasks, the colors are taken from a specific list. If there are more than 8 tasks, then the missing colors are generated randomly.

pie_charts_with_small_number_of_tasks

pie_charts_with_lot_of_tasks

view_all_tasks_modal


EvianTan:
All changes verified!! 👍 
The pie chart part is awesome and beautiful!

EvianTan (approved):

src/components/common/PieChart/PieChart.jsx
```diff
@@ -0,0 +1,64 @@
+import { useEffect } from 'react';
+import * as d3 from 'd3/dist/d3.min';
+import { CHART_RADIUS, CHART_SIZE } from './constants';
+import { generateArrayOfUniqColors } from './colorsGenerator';
+import './PieChart.css';
+
+export const PieChart = ({ data, dataLegend, pieChartId, dataLegendHeader }) => {
```
andrewpaulino, line 7: Hi! One random recommendation here is to try to memoize this component if you can, when react reconciles it may cause this to render again from the parent. Small optimization, but could be good to have in the future if the parents cause re-renders, because basically it will keep re-drawing the pie chart.

veratim1996: I think, at the moment this optimization is not helpful. Memoization itself requires memory, so I don't see the point in optimizing prematurely.

PR #506, opened 2022-09-09
Team task UI change

Programmingrelated1999:
This PR is for making the “Team Member Tasks” section a tab as part of the “Time Entries” section.

There were several changes made. They are listed below. 

* The TeamMemberTasks is no longer rendered from the Dashboard component, instead it is now rendered as a tab in Timelog component, and is passed asUser variable which is also passed as props.asUser from Dashboard to Timelog. 
* activeTab was used to be assigned as 0 for current week timelog, 1 for last week, 2 for week before last, and 3 for search by Date Range. Now it is assigned to as 0 for tasks, 1 for current week timelog, 2 for last week, 3 for week before last, and 4 for search by Date Range.
*  The activeTab was used in startOfWeek() and endOfWeek() for calculating dates for the week. Since the value of activeTab changes now, i use activeTab - 1 to adjust the changes. For example value for current week timelog now is 1, while it used to be 0 so to adjust it subtract 1. 
* activeTab condition in EfforBar.jsx is also changed.

Now the time Entries has one more tab which is Tasks. The summary also shows the same way as previously. Current Tab text color is also black while other tab text color is blue. 

The screenshots are shown below:
TeamTaskTimeLog1
TeamTaskTimeLog2

PR #518, opened 2022-09-17
Move pseron task if assigned

Programmingrelated1999:
Main function is to move current user to top of the Tasks list if there are any tasks assigned. This is also fixed and built from pull request https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/506. 

A few Changes to note: 
* To show current user to top of the Tasks list if there are any tasks assigned. I first get currentUser object from the list for comparison. Then a conditional variable is created to determined if there are any tasks assigned.
* I loop through the current user's tasks to see if there are any tasks attached to a projectId and taskId. If there is one, the conditional variable is set to true. 
* Based on the true/false value, the current user will be removed from the list of persons and shifted to the top of the list.
* It displays "Tasks and Timelogs" instead of "Time Entries".
 
code_preview
TeamMemberTasks_UI_movecurrentUserPriority
TeamMemberTasks_UI_movecurrentUserPriority2


src/components/TeamMemberTasks/TeamMemberTasks.jsx
```diff
@@ -122,68 +144,76 @@ const TeamMemberTasks = (props) => {
...
-                                }}
-                                />
-                              }
-                          </p>
-                        </td>
-                        {task.hoursLogged != null && task.estimatedHours != null &&
-                          <td className='team-task-progress'>
-                            <div>
-                              <span>
-                                {`${parseFloat(task.hoursLogged.toFixed(2))}
```
brucelin006, line 152: Hi Kaung, I noticed that this orange part seems to be exactly the same as the blue part below except that the blue part moves `(task, index) =>` to a new line, is this done by the code formatter?

brucelin006 (commented):
Hi Kaung, I think everything is good to go, but could you please take a look at my comment on the second file, because I am not sure if this is a code style issue.

src/components/TeamMemberTasks/TeamMemberTasks.jsx
Programmingrelated1999: Hi Bruce, yes I believe this is done by the code formatter.

PR #519, opened 2022-09-24
Move pseron task if assigned

Programmingrelated1999:
Main function is to move current user to top of the Tasks list if there are any tasks assigned. This is also fixed and built from pull request https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/506. 

A few Changes to note: 
* To show current user to top of the Tasks list if there are any tasks assigned. I first get currentUser object from the list for comparison. Then a conditional variable is created to determined if there are any tasks assigned.
* I loop through the current user's tasks to see if there are any tasks attached to a projectId and taskId. If there is one, the conditional variable is set to true. 
* Based on the true/false value, the current user will be removed from the list of persons and shifted to the top of the list.
* It displays "Tasks and Timelogs" instead of "Time Entries".

Additional functionality of Tasks:
* Other Members are sorted in alphabetical order.
 
code_preview

TeamMemberTasks_UI_movecurrentUserPriority
Sorted


brucelin006 (approved):
Everything looks good.

PR #521, opened 2022-09-28
update files in /src/actions

EvianTan:
This PR is **first part** of "enabling ESLint for frontend Repository".

Changes in this PR:
1. Change the `.eslintignore` file
So ESLint works for all "*.js" and "*.jsx" files in this repository, instead of the `/src/__tests__/*` directory

2. Update the `.eslintrc.js` file
Add so many rules to remove the red underlines of our codes. If you're interested what are they, please check [here](https://eslint.org/docs/latest/rules/). Or if you have different options like which rules should be warn, which rules should be off, feel free to let me know 😄 

3. Refactor files in `/src/actions` per ESLint rules, includes:
- quotes problems
- indents problems
- import order problems
- import path problems
- use `let`, `const` instead of `var`
- remove unused comments
- remove unused printouts
- string concatenation problems
- export default errors
- etc.

This PR shouldn't do any functional changes. If you find anything break, just let me know.
I'll continue refactoring all the remaining JS files in the repo in next coming PR.


kseniaxu0704 (approved):
Thanks for the pr, LGTM, approved.

PR #525, opened 2022-10-05
add condition for non tasks for current user

EvianTan:
This PR is a fixing for the bug that "non Admin/Owner users cannot login on DEV/local successfully".

The below image is the problem description:
image

I found that problem came from the `src/components/TeamMemberTasks/TeamMemberTasks.jsx` file. If the the current user doesn't have any tasks, it would hit an error there. So I just add a condition to tell if the `currentUser` existed (new line 89).

**How to test it:**
1. check into my branch
2. test with login/logout with your admin/manager/volunteer accounts


Red4lpha:
Tested with this branch - It is now working for me, no matter if the account is an admin or not.

Programmingrelated1999 (approved):
Everything looks good on my computer for all different roles

rubastricks:
it's working fine in my local

PR #526, opened 2022-10-06
Timelog shows correct time and week before last works

Programmingrelated1999:
Hi,

I realize that I forget to adjust the data transferred to Effort Bar needs to be corrected. Previously there are 4 tabs, but however there are 5 tabs now and the first tab(tab 0) is the Tasks tab. Therefore all other tabs 1,2,3,4 used to be tabs 0,1,2,3. To adjust the change, I subtract -1 in timeEntries week as state.timeEntries.weeks[activeTab -1].

The week before last seems to be working now but please let me know if you can click on week before last with no errors
Capture


aashish-thapa-magar (commented):
I ran your code on my local. It no longer shows a blank page.

rubastricks:
no longer showing a blank page

aashish-thapa-magar (approved):

PR #530, opened 2022-10-13
Fixed: total badge count + header data issue when switching users

aashish-thapa-magar:
Hello,
I have fixed the total badge count and the header data issue where data is one refresh behind after a new user login. But, I am not sure if it's a proper fix. I will try to add details in the comments.

Please provide suggestions, if any

### DIFF
--- a/src/components/Badge/Badge.jsx
+++ b/src/components/Badge/Badge.jsx
@@ -17,7 +17,7 @@ import OldBadges from './OldBadges';
 import BadgeReport from './BadgeReport';
 import { getUserProfile } from '../../actions/userProfile';
 
-const Badge = (props) => {
+const Badge = props => {
   const [isOpen, setOpen] = useState(false);
   const [isOpenTypes, setOpenTypes] = useState(false);
   const [totalBadge, setTotalBadge] = useState(0);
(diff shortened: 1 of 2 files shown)


src/components/Header/Header.jsx
```diff
@@ -45,10 +45,15 @@ export const Header = props => {
...
+      console.log(
+        'header.js: props.auth.isAuthenticated',
+        props.auth.isAuthenticated,
+        props.auth.user.userid,
+      );
       props.getHeaderData(props.auth.user.userid);
       props.getTimerData(props.auth.user.userid);
     }
-  }, []);
+  }, [props.auth.isAuthenticated]);
```
aashish-thapa-magar, line 56: checks if a user is authenticated and loads data accordingly

src/components/Badge/Badge.jsx
```diff
@@ -17,46 +17,18 @@ import OldBadges from './OldBadges';
...
-    setOpen((isOpen) => !isOpen);
-  };
-
-  const toggleTypes = () => {
-    setOpenTypes((isOpenTypes) => !isOpenTypes);
-  };
-
-  useEffect(() => {
-    const userId = props.userId;
-    props.getUserProfile(userId).then(() => {
```
aashish-thapa-magar, line 55: I removed this because I don't see any association from this code. Also, it's the main reason behind accurate data being one refresh behind after login

src/components/Badge/Badge.jsx
```diff
@@ -65,8 +37,31 @@ const Badge = (props) => {
...
+        console.log('badge2', badge);
+        if (badge?.badge?.badgeName === 'Personal Max' || badge?.badge?.type === 'Personal Max') {
+          count += 1;
+        } else {
+          count += badge.count;
+        }
+      });
+      setTotalBadge(Math.round(count));
+    }
+  }, [props.userProfile.badgeCollection, totalBadge]);
```
aashish-thapa-magar, line 64: this re-renders the component as userProfile isn't immediately updated after login which is why I am not sure if this is a proper fix. Also, the two state changes re-render the badge components multiple times.

guiwustro (changes requested):
I did some tests here, and the badges still not updating the total number properly  I mean, the total of badges must be the count sum, right? In this case should be 14 and It's appearing 8.  I take a look in the logic and I think a good approach would be doing a reduce to sum the count in the badge object. I sent some printscreens in the slack

guiwustro (approved):

PR #531, opened 2022-10-14
Fixed to badges not updating immediately on the user profile screen when deleting or assigned badges.

Red4lpha:
Per bug request - fixed:
a: Badge doesn’t actually delete.

- On the Main it deleted, then came back. I was able to make them come back by using the “Assign Badges” button and assigning a badge. Then going back into the “Select Featured” and there all the deleted ones were.  
b: If I delete a badge that is checked, then try and select a new badge, I get the warning that says I can only select 5 badges 
c: Assigning badges also doesn’t save: Profile → Featured Badges → Assign Badges → Select → Save Changes_

- Badge list doesn’t update

### DIFF
--- a/src/components/Badge/BadgeReport.jsx
+++ b/src/components/Badge/BadgeReport.jsx
@@ -224,13 +224,18 @@ const BadgeReport = (props) => {
   };
 
   const saveChanges = async () => {
-    let newBadgeCollection = sortBadges.slice();
+    let newBadgeCollection = JSON.parse(JSON.stringify(sortBadges));  
     for (let i = 0; i < newBadgeCollection.length; i++) {
       newBadgeCollection[i].badge = newBadgeCollection[i].badge._id;
-    }
+    }  
     console.log(newBadgeCollection);
     await props.changeBadgesByUserID(props.userId, newBadgeCollection);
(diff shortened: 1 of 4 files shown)


src/components/Badge/BadgeReport.jsx
```diff
@@ -217,20 +217,28 @@ const BadgeReport = (props) => {
...
-    }
+    }  
     console.log(newBadgeCollection);
     await props.changeBadgesByUserID(props.userId, newBadgeCollection);
     await props.getUserProfile(props.userId);
+    
+    props.setUserProfile(prevProfile => {
+      return {...prevProfile, badgeCollection: sortBadges }
+    });
+    props.handleSubmit();
```
Red4lpha, line 241: This is a function passed from UserProfile.jsx that is used for when 'saving' the user profile. Using setUserProfile wasn't enough alone to stop old stale data from coming back after manually refreshing the browser. I know we're trying to move a lot of functions from UserProfile to get it from being a 'monolith', but I doubt this important centralized function will be moved. Let me know if this is not the case.

EvianTan:
Changes verified. Great job!

EvianTan (approved):

PR #533, opened 2022-10-14
Frontend: Fix projects pie chart

veratim1996:
There is a problem: the names of the projects that the user's tasks belong to are taken from the wrong place. And because of this, sometimes project names are not displayed. Now, data about projects is taken from userTask.

image


EvianTan:
Code looks great and test with no weird things. The only concern is I cannot find or reproduce the problem in description.

EvianTan (approved):

PR #534, opened 2022-10-15
Redesign the project report page

veratim1996:
Project report page redesign.

 **New look**:

final-1
final-2

**As it was:** 

old


Red4lpha (approved):
I double checked this along with Kaung Htet Myat. Appears correctly,

PR #542, opened 2022-10-26
Fix to featured checkbox not aligning properly in the featured badges area

Red4lpha:
Badges bugs 1. a - Fix requested by Jae

### DIFF
--- a/src/components/Badge/BadgeReport.jsx
+++ b/src/components/Badge/BadgeReport.jsx
@@ -13,6 +13,7 @@ import {
   Modal,
   ModalBody,
   ModalFooter,
+  FormGroup
 } from 'reactstrap';
 import pdfMake from 'pdfmake/build/pdfmake';
 import pdfFonts from 'pdfmake/build/vfs_fonts';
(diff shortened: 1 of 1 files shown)


src/components/Badge/BadgeReport.jsx
```diff
@@ -319,14 +320,19 @@ const BadgeReport = (props) => {
...
                 <td style={{ textAlign: 'center' }}>
-                  <Input
-                    type="checkbox"
-                    id={value.badge._id}
-                    checked={value.featured}
-                    onChange={(e) => {
-                      featuredChange(value, index, e);
-                    }}
-                  />
+                  <FormGroup check inline >
```
Red4lpha, line 323: So the issue here was that reactstrap was placing a placing an absolute positioning on the checkbox input for some reason. So as noted in the comment, this can also be fixed by overriding the position style within the input: `style={{ position: 'static' }} `
Though, that felt a bit hacky to me, so I opted to use something in reactstrap. Though this does not look at clean.

talls2:
Nice, less code is always better

talls2 (approved):
Nice, less code is always better.

PR #547, opened 2022-11-05
Kaixiang fix volunteering time

KaixiangGU:
This PR is about Fixing volunteering hours by category. Changes were made in this PR:

1. Added three new categories: Total Intangible Hours, Total Economics Hours, and Total Stewardship Hours.
2. Total Intangible Hours should pull back all your past Intangible hours.
3. All those Category hours should be increased based on your selected project in your time entry, as well as "Total Tangible Hours" should be increased at the same time.
4. When someone's hours are manually edited, new hours should be added on to them and hours should keep tracking afterwards. 
5. When an old time entry was edited, especially volunteering times were edited, category hours and total Total Tangible Hours should increase/decrease depending on your old hours.
6. All hours should be displayed in two decimal places.


src/actions/timeEntries.js
```diff
@@ -73,7 +73,7 @@ export const editTimeEntry = (timeEntryId, timeEntry, oldDateOfWork) => {
     try {
       const res = await axios.put(url, timeEntry);
       dispatch(updateTimeEntries(timeEntry, oldDateOfWork));
-      return res.status;
+      return res.status
```
LeeSingChanYau, line 76: Did you delete the semicolon on purpose?

src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
```diff
@@ -75,13 +75,14 @@ const TimeEntryForm = (props) => {
...
   const [tasks, setTasks] = useState([]);
+  const [formDataBeforeEdit, setFormDataBeforeEdit] = useState({});
 
   const fromTimer = !_.isEmpty(timer);
   const { userProfile, currentUserRole } = useSelector(getTimeEntryFormData);
 
   const dispatch = useDispatch();
 
-  const tangibleInfoToggle = (e) => {
+  const tangibleInfoToggle = e => {
```
LeeSingChanYau, line 85: Did you delete the parenthesis in line 85 and 47 to be consistent in how you write the code?

src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
```diff
@@ -112,29 +113,38 @@ const TimeEntryForm = (props) => {
...
       })
       .catch(err => console.log(err));
   }, []);
 
+  //grab form data before editing
+  useEffect(() => {
+    if (isOpen && edit) {
+      setFormDataBeforeEdit(inputs);
+    } else {
+      setFormDataBeforeEdit({});
```
LeeSingChanYau, line 127: is it necessary to set setFormDataBefore to empty object every time that isOpen or edit is false?

LeeSingChanYau (commented):
I do not see any problems with the code, just have some doubts about the syntax.

src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
KaixiangGU: I think this is caused by the prettier auto re-format the file when I changed the code on this file.

src/components/Timelog/TimeEntryForm/TimeEntryForm.jsx
KaixiangGU: Thanks for pointing it out, it's not necessary to write else condition, I'm gonna change it.

src/actions/timeEntries.js
KaixiangGU: Hmm, no, maybe I accidentally deleted it.

guiwustro:
Its working well here. The only issue that I found is that I need to refresh the page to see the changes.

guiwustro (approved):
Its working well here. The only issue that I found is that I need to refresh the page to see the changes. I uploaded a video in the thread of the PR at slack

PR #548, opened 2022-11-05
Permissions dashboard for an owner

guiwustro:
The owner should be:
- Change the permission of all existing roles;
- Create a new role and choose permissions for the role;
- Assign a role for any user in the UserManagement;
- Assign "Specific" permissions for a user;
userRolesLayout
manageUserPermissionsOtherPerson
PermissionsManagement-Modal-1

- The docs in the drive, show some about what was done.
P.S: If a permission of a logged user is changed, the changes are not going to happen if the user doesnt logout and login again, changing the token.
https://docs.google.com/document/d/1rzCvlu5FzIXoUpWIM07Z5g80UgqN2DikT2Csz1FdtV8/edit


KaixiangGU (approved):
I tested with an owner account, everything works well.

PR #553, opened 2022-11-10
Fixed infringment mispelling to infringement

LeeSingChanYau:
Lots of misspelling of “Infringment” as the variable name in the current code base. Fixed the misspelling in each file and changed filenames that were misspelled.


src/components/RenderInfringement/RenderInfringment.jsx
```diff
@@ -4,11 +4,11 @@
 // import _ from 'lodash'
 // import ShowSaveWarning from "../common/ShowSaveWarning"
 
-// class RenderInfringment extends Form {
+// class Renderinfringement extends Form {
```
KaixiangGU, line 7: There are multiple names not using camelCase in this file. 
This file name is also misspelled.
Please check it, thank you!

LeeSingChanYau: Thanks for the catch. Fixed!

KaixiangGU (approved):
All Good!

PR #559, opened 2022-11-16
Rework on the WBS Task actions and fixing task deletion

Red4lpha:
Originally my task was to implement a delete button for deleting tasks on the WBS task table. After diving into the code I noticed there was already functionality and UI for this, it was just hidden and didn't work very well.

After talking to Jae, my task changed to:

> It could even still say “edit” but include the little arrow so it’s clear that you are entering an “edit” process and then the arrow reveals all the options, including the delete option and any others we want to add (like “remove resources”). I’d really like to be able to hide those with the same button/arrow too.
> 
> As a matter of fact, I just saw on the far right that a “remove” button already exists! Please just move that over with the other buttons and then implement what we’re talking about here and I think we can mark this action item as done!

So, with this PR, the changes include:

1. Fixing the 'Remove' to actually delete the tasks
2. Changed the 'Edit' button on tasks to now display the hidden secondary table row with all the available task actions. This is toggleable, so it will toggle back and forth.
3. Changed to Remove button to no longer float right away from the other buttons as per Jae's request.

[Image of new UI](https://www.dropbox.com/s/aoc3yv8c0x2q76c/Week7-WBSTask-NewUI.png?dl=0)


EvianTan:
Great job @Red4lpha  !! I really like the toggle "Edit" button.

✅ Verified changes:
1. Move the extra action popup under the new "Edit" button
2. Improve the popup UI, move the "Remove" button to the left
3. Make the "remove" button work successfully

💡 The thing can be improved:
The extra action popup could take the whole row, currently it misses the last one column. (see below image)
image

Red4lpha:
- I changed the controller row to dynamically change colSpan based off it's parent. So now if we add or remove columns it will always match it's parent.
- Ran prettier on the file to clean up formatting.

EvianTan (approved):

one-community:
Looks good. 

1) Please fix the popup formatting to: 

Hold up there pally, you sure you want to delete this!?!
 
Deleting this cannot be undone and the process deletes all the tasks within this folder too. If you want to keep those and still delete this folder, you first need to drag the ones you want to keep to another folder.

Easy peasy, then you can destroy this folder forever and feel good about it.

2) Please confirm that deleting these does not delete their associated data... We need to still have hours and other details available for when we run project and people reports. 

3) The info in #1 pops up when deleting individual TASKS too. That popup should just be for FOLDERS. Please create this popup instead for TASKS: 

Whoa tiger, you sure you want to delete this!?!
 
Deleting this cannot be undone and removes the item from this WBS and the Tasks list of anyone listed on it as a Resource. The data remains for people, project, etc. reports functions but that's it! If this WBS will be used in the future by others AND this task is one that is important to still be a part of this WBS for them (or any other reason you can think of), consider just marking it "Complete" instead: Edit --> Status --> Complete --> Save

If this task was a "one-off" or other non-useful WBS task, then delete with confidence, nobody wants to see it here anymore anyway.

one-community:
I've merged this but am not seeing changes like the little dropdown arrow on the Main... Any idea why this might be? 
image

PR #563, opened 2022-11-18
deleted yarn

LeeSingChanYau:
Remove the yarn files such as yarn.lock from the repo keeping only npm as the package manager.


KaixiangGU (approved):
Approved

PR #235, opened 2021-08-24
Timer button fix2

CameronWhiteCS:
Solves issue:

> Jae: Dashboard → Timer Appearance: Please center the word “Pause”. “Clear” looks like it may be off center too.

PR #580, opened 2022-12-02
fix userProfile not loading

KaixiangGU:
UserProfile should work now...

I would say this is a temporary fix to let us load userProfile. 
The reason behind this bug is all the current users have been already created with "infringments" property, but since we've changed all the misspellings of "infringments" to "infringements", it WILL NOT load blue squares correctly on user profile page. 

And notice that once the **[Backend PR](https://github.com/OneCommunityGlobal/HGNRest/pull/220)** of misspellings of "infringments" which Alan is working with is merged, all the new users will be created with "infringements" property. So, for now I just keep both "infringments" and "infringements" on the file `BlueSquareLayout.jsx`.

What I thought is we might experience new bugs in the future when we handle blue squares of users because of the terminology discrepancies. Maybe the best solution for this is to find a way to rename all the "infringments" property to "infringements" in the database. 

Hope this makes sense!

user_profile_page

### DIFF
--- a/src/components/UserProfile/BlueSquareLayout.jsx
+++ b/src/components/UserProfile/BlueSquareLayout.jsx
@@ -6,9 +6,18 @@ import './UserProfile.scss';
 import './UserProfileEdit/UserProfileEdit.scss';
 
 const BlueSquareLayout = props => {
-  const { userProfile, handleUserProfile, handleBlueSquare, isUserSelf, role, canEdit } = props;
+  const {
+    userProfile,
+    handleUserProfile,
+    handleBlueSquare,
+    isUserSelf,
+    role,
+    roles,
+    userPermissions,
+    canEdit,
+  } = props;
 
-  const { privacySettings, infringements } = userProfile;
+  const { privacySettings } = userProfile;
 
   if (canEdit) {
     return (
@@ -26,7 +35,7 @@ const BlueSquareLayout = props => {
         </div>
 
         <BlueSquare
-          blueSquares={infringements}
+          blueSquares={userProfile?.infringments || userProfile?.infringements}
           handleBlueSquare={handleBlueSquare}
           role={role}
           roles={roles}
@@ -42,7 +51,11 @@ const BlueSquareLayout = props => {
       ) : (
         <div>
           <p>BLUE SQUARES</p>
-          <BlueSquare blueSquares={infringements} handleBlueSquare={handleBlueSquare} role={role} />
+          <BlueSquare
+            blueSquares={userProfile?.infringments || userProfile?.infringments}
+            handleBlueSquare={handleBlueSquare}
+            role={role}
+          />
         </div>
       )}
     </div>


one-community:
Thank you for this. My understanding is, once Alan's backend PR is merged, that should update all those misspellings in the backend too.

src/components/UserProfile/BlueSquareLayout.jsx
```diff
@@ -26,7 +35,7 @@ const BlueSquareLayout = props => {
         </div>
 
         <BlueSquare
-          blueSquares={infringements}
+          blueSquares={userProfile?.infringments || userProfile?.infringements}
```
LeeSingChanYau, line 38: Looks good as a temporary solution. We will try to fix the misspellings from here to make sure it works.

src/components/UserProfile/BlueSquareLayout.jsx
```diff
@@ -42,7 +51,11 @@ const BlueSquareLayout = props => {
       ) : (
         <div>
           <p>BLUE SQUARES</p>
-          <BlueSquare blueSquares={infringements} handleBlueSquare={handleBlueSquare} role={role} />
+          <BlueSquare
+            blueSquares={userProfile?.infringments || userProfile?.infringments}
```
LeeSingChanYau, line 55: Same as above. We just need to keep in mind that we will need change the spelling at some point.

LeeSingChanYau (approved):
Looks good!

EvianTan:
> Thank you for this. My understanding is, once Alan's backend PR is merged, that should update all those misspellings in the backend too.

I don't think so. I just checked Alan's backend PR, it changed all misspell about infringement, but that PR cannot update the existing records in our database. That would only affects on new users' infringements.

one-community:
Oh! Ok, we need to update that for past users too or things will be broken, right? Is there a way to easily do that for all past/existing users? If it difficult to do for all past users, can we just do it for the active 33?

KaixiangGU:
That's right. But blue squares are populated by infringements property in **userProfile object**. And all the users were already created with old terminology "infringments" in the database, I think that WILL NOT change when Alan's backend PR is merged, and new users will be created with new terminology "infringements" after that point. So that's why I said we may have new bug when dealing with blues quares because the property name is different. 

image

LeeSingChanYau:
I just spoke to Kaixiang and I believe that we should keep BlueSquareLayout as it is now after merging this PR, and it should be safe to update other misspellings of "infringement" on other files.

KaixiangGU:
Maybe find a way to rename that "**infringments**" property to "**infringements**" for all the users in database? Will be the best solution?

one-community:
> Maybe find a way to rename that "**infringments**" property to "**infringements**" for all the users in database? Will be the best solution?

Yes, this seems like it would fix the problem most comprehensively, right?

KaixiangGU:
Yes, at least THIS PR is to make sure we can load the user profile for now, otherwise, we can't do anything without that.

one-community:
Ok, sounds good. Merged this and the other one and everything is working now on the Dev. Merging to Main/Production. Thank you all!

one-community:
Can someone review PR #235 also? That's another really important one.

PR #581, opened 2022-12-02
View other people dashboard popup fixed.

LeeSingChanYau:
This issues solves two issues when viewing someone else's dashboard. The first one was that the popUpBar was not staying fixed to the top when we scroll through the page, now it sticks to the top no matter where we are. The second one was sometimes the tasks showed up on the Timelog component, I made sure that the tasks loaded are based on the prop passed from the dashboard component and not the current user logged in. This works in my end but requires more testing.


ShaunSully88 (dismissed):
Awesome! Good eye capitalizing those rogue words...

EvianTan:
Hey @LeeSingChanYau ,

Sorry I couldn't give a approval on this PR for couple reasons below:

1. This PR contains a lot ESLint/formatting changes unrelated the description. **One PR should only do one thing.** Plus with your changes, there're lots of files will full of wave lines, it would be harder for people to debug or develop.
2. The description on this PR is not clear enough. Please add more images like before/after comparing.

I would suggest you to close this PR, create a new PR with only the above two issues' fixing. Add more detailed description and images as before/after changes. You can copy, paste and modify this Markdown template as your PR description.  
image

LeeSingChanYau:
Hello @EvianTan, yes those other changes in ESlint were not meant to be in this branch. I was careless not to check before pushing. I will close this PR and make a new one shortly with just the necessary changes. The problem described above was about my previous PR where we click the dot next to a person's name and a new Dashboard window will pop up. The wrong information sometimes will show up in the Timelog component. 
Screenshot 2022-12-15 at 9 35 41 AM

PR #584, opened 2022-12-03
Yiyun hotfix volunteer cannot login

EvianTan:
**The bug is: volunteer users cannot login on DEV and local.**

image (1)

**The reason is the role of users should be `Volunteer`, but somehow from frontend it shows `Volunteer Leader` instead.**
page emtpy reason
So the `roleIndex` = -1 and `permissions` is undefined, that causes the render error.
image

This PR add a check for `roleIndex` to avoid issue.

**Things to do after this PR:**
Figure out the root cause why the user role is `Volunteer Leader`.


KaixiangGU (approved):
Looks good! Approved

PR #587, opened 2022-12-05
Fix Summary submitted

LeeSingChanYau:
Updated getWeeklySummary function in SummaryBar.jsx to check if the time is less than the last summary due date.

### DIFF
--- a/src/components/SummaryBar/SummaryBar.jsx
+++ b/src/components/SummaryBar/SummaryBar.jsx
@@ -211,7 +211,16 @@ const SummaryBar = props => {
 
   const getWeeklySummary = user => {
     let summaries = user.weeklySummaries;
-    if (summaries && Array.isArray(summaries) && summaries[0] && summaries[0].summary) {
+    const timeNow = new Date();
+    const latestSummaryDueDate = new Date(summaries[0].dueDate);
+
+    if (
+      summaries &&
+      Array.isArray(summaries) &&
+      summaries[0] &&
+      summaries[0].summary &&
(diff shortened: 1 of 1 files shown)


one-community:
Helping someone who finishes their hours on Sundays submit their summary just now, the app also seems to be auto-filling the summary part with last week’s summary… which may be what is causing the app to think they’ve already been submitted. So this way of fixing it might not be the actual source of the problem.

LeeSingChanYau:
Ok, I think this is a backend problem then because I was testing earlier and I hardcoded the summaries array where the most recent one was empty but the second one had some text and its behaving correctly.

one-community:
I think deleting the old summary from the summary entry field to enter the current/new summary also deletes it from the "last week" tab too... Then entering the new summary enters it in for this week. But last week's is now missing (when I run a Summaries Report for the whole team for the week) and entering it in for last week and saving it doesn't add it to the report but DOES save it in the summary entry field.

johny4437 (approved):

EvianTan (approved):

EvianTan:
Approved with creating a new User, the Wkly Summary shows correct.

one-community:
This worked for me but got red notice on right saying "data could not be saved". Data did save though.... merging to Main and we'll see what this does there.

one-community:
also had to refresh page to get green checkmark to show up at top... page should auto-refresh to show that... will add as new bug after merge to Main

one-community:
Great fix. Final steps needed here have been added to the Bugs Doc.

PR #592, opened 2022-12-10
auto refresh the page when after clicking pause and resume button

kseniaxu0704:
When the user click the pause and resume button, it is possible to auto-refresh page after 5 second select reactivate date and delete reactivate date


KaixiangGU (approved):
Tested! It works well 👍

PR #596, opened 2022-12-15
Fix the blue squares cannot be edited/assigned/deleted problem

EvianTan:
# Description
Urgent bug #4
image
✅ This PR will solve the "blue squares cannot be edited/assigned/deleted" problem
❓ Not sure if it can also solve "add blue squares should send an email" problem 

## Mainly changes explained:
Before changes, `userProfile?.infringments || userProfile?.infringements` will only show all `infringments` instead of `infringements ` if it is not null, so `infringements` will be totally ignored.

## How did I find it:
1. I tried to add blue squares on DEV, it did save newly added blue squares on database. See blow image as an example
blue squares can be added but not show up

2. I found the frontend props of useProfile also contained the blue squares I added
4. I found the blueSquareLayout only show old `infringments` data and ignore all new `infringements` data
5. I found `userProfile?.infringments || userProfile?.infringements` will only show all `infringments` instead of `infringements ` if it is not null, so `infringements` will be totally ignored.

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. log as admin/owner user
4. go to a user's profile page
5. try edit/assign/remove blue squares

## Note:
1. ~~To get rid of future bugs, I'll move the previous `infringments` records under `infringements` for all active users on our BETA database.~~ Already modified on BETA database.
2. When you test, please ignore the below popup error. It already files as urgent bug 10 and not introduced by this PR.
image

PR #597, opened 2022-12-16
fix undefined errors when rendering user profile page

EvianTan:
# Description
Urgent bug #11
image

## Mainly changes explained:
Changes based on the terminal output

## How to test:
check into current branch
do `npm install` and `...` to run this PR locally
log as volunteer user
add Jaeaa Test into your current team
verify his profile page can be accessed now

## Note:
The terminal output about blue squares only happened on Main currently, might because of I already updated Main DB but my last PR #596 didn't merge into Main.


RJBornToWin (approved):
I approve the changes.
Tested on local copy with admin and volunteer account. Able to access other user profiles.

NOTE: This fixes the UI blank page error as a work around and doesn't address the root cause for the issue. The same bug can be tested later when all the major bug fixes are complete and pushed to main.

PR #600, opened 2022-12-17
fix cannot render blue squares

EvianTan:
# Description
image

## Mainly changes explained:
Cannot reproduce on DEV, so just caught all the errors I can hit on Main when click on the blue squares
main blue square cannot render error 1
main blue square cannot render error 2

## How to test:
For Dev, test and don't break anything
For Main, no way to test


KaixiangGU (approved):
Works well on my local. Let Jae test it on the main.

PR #603, opened 2022-12-18
Feat/show assigned tasks on user profile

guiwustro:
# Description
This PR will change the way the tasks are deleted from the UserProfile dashboard. Now, when you delete, the task is removed from the user. The new behavior will just be filtering what is shown in each userProfile and in the timelog. I believe, this task is important to make it easy to manage the volunteers that have a lot of tasks, and in this way, the tasks of each volunteer could be managed by the userprofile.

## Mainly changes explained:

- The projects tab in the userProfile, in the Task Table at the first render will just show de Active tasks from the user (that means, the tasks that is not completed)
- The task table will have buttons to filter the active, completed and all tasks related to this user.
If the task is completed, will have a button to transform this task to active, and if the task is active, has a button named "delete" to change to the complete.
- The timelog component was rendering all tasks from the user, now will just render the active tasks.

## How to test:
- You need to test using the branch feat/Guilherme-FixDeleteTasksFromUserProfile (back-end):
yarn build
yarn start
- And then, execute in the branch feat/showAssignedTasksOnUserProfile

1. Log as admin/owner user
Now, to check the feature, you should enter in some userProfile that has tasks, to do this, I put bellow how it can be done:
A. go to OtherLinks -> Projects -> Click in some WBS project name that has tasks (the first one preferably) -> going to some URL like:
.../wbs/tasks/627071763a043a347c8d8c65/5f6cdff77b73161ddb32a0e5/Jae's%20WBS%20-%20USE%20THIS
B. In the tasks you should click to some member, like in the screenshot below:
https://prnt.sc/y00czuiaU-CU
C. Inside the profile, you should go to the tab "Projects", and there you will be able to check the feature.
--
To check the feature, can be done tests like:
1. Delete/Add Tasks inside the projects tab and save to check if is working;
2. In the clock of each userProfile, beside the name, you also can check the tasks list, if the tasks are being listed correctly (the tasks that need to be show is just the actives tasks)
3. You can also check if deleting the task will be changing the tasks dashboard (but I'm sure that will not be changing because should not)

--- Test 2:
1. log as a volunteer (that has tasks in the profile)
2. Check if the changes made by an admin is taking effect (To do this you should assign tasks for this user through the tasks dashboard before login here)
Note (tip): If you are testing this, one tip is to log as admin in the browser and in an anonymous browser you can log as a volunteer (is what I usually do), this way you can check in both at the same time.

-Some pictures about the new project tab in user Profile: 
 View of a owner/admin (person that has permissions to delete/add active tasks):
ActiveTasks

View of a volunteer
UserProfileProjectTabOfAVolunteer

### DIFF
--- a/src/components/TeamMemberTasks/TeamMemberTasks.jsx
+++ b/src/components/TeamMemberTasks/TeamMemberTasks.jsx
@@ -169,9 +169,14 @@ const TeamMemberTasks = props => {
                 <tbody>
                   {user.tasks &&
                     user.tasks.map(
-                      (task, index) =>
-                        task.wbsId &&
-                        task.projectId && (
+                      (task, index) =>{
+                        let isActiveTaskForUser = true
+                        if (task?.resources) {
+                          isActiveTaskForUser = !task.resources?.find(resource => resource.userID === user.personId).completedTask
+                        }
+                        if (task.wbsId &&
+                        task.projectId && isActiveTaskForUser ) {
+                          return (
                           <tr key={`${task._id}${index}`} className="task-break">
                             <td className="task-align">
                               <p>
@@ -213,8 +218,8 @@ const TeamMemberTasks = props => {
                               </td>
                             )}
                           </tr>
-                        ),
-                    )}
+                        )}
+                                  })}
                 </tbody>
               </Table>
             </td>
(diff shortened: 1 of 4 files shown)


EvianTan:
Hi @guiwustro ,

Fantastic job on this feature! I also like your description above! I can verify all the changes and features.
Like I said, the only reason I didn't approve this PR is it's related to your backend PR. Once you update the backend one, I'll also approve this one.!

EvianTan (approved):

PR #614, opened 2022-12-22
hot fix blue squares cannot be deleted on Main

EvianTan:
# Description
image

## Mainly changes explained:
image
I was thinking, the delete action doesn't work on Main, because of line 269 get `infringements` as an empty array first, so line 270 `infrigement._id` is undefined, so that `newInfringements` has all items from `infringements`. That turns out no items are deleted. 

## How to test:
Since I cannot reproduce them on Dev, I cannot guarantee they are fixed on Beta. 

## Note:
This hotfix PR just like #600


KaixiangGU (approved):
Let's try this PR first.

If this one does not work, I think another reason could be some of the blue squares were created with "infringments" property instead of "infringements".

PR #615, opened 2022-12-22
Added in regions section along with enableLongNumbers

ShaunSully88:
BUG:  (WIP SHAUN) App not allowing me to create new users with foreign phone numbers
Other Links → User Management → Create New User → input foreign number like the example below → Gives error shown below and won’t allow me to save

FIX: Added in "enableLongNumbers" and "regions" keys to the PhoneInput section, all international numbers should now be able to be entered without issue

### DIFF
--- a/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
+++ b/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
@@ -176,8 +176,7 @@ class AddUserProfile extends Component {
                   <Col md="6">
                     <FormGroup>
                     <PhoneInput
-                        country="US"
-                        enableLongNumbers = 'true'
+                        country="US
                         regions={['america','europe','asia','oceania','africa']}
                         value={phoneNumber}
                         onChange={phone => this.phoneChange(phone)}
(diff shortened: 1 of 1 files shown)


src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
```diff
@@ -176,8 +175,10 @@ class AddUserProfile extends Component {
                   </Col>
                   <Col md="6">
                     <FormGroup>
-                      <PhoneInput
-                        country={'us'}
+                    <PhoneInput
+                        country="US"
+                        enableLongNumbers = 'true'
```
navyasrim, line 180: if we add this it will not restrict the length for countries.. like US need to allow 10 but now it is allowing 20..not showing validation also..

ShaunSully88: So i just tried adding in numbers for both Canada and United States, and it did limit to 10 each. So I do think it is limiting based on region selected? I am not familiar with international numbers but if someone wanted to try one respective to a country/region where they're aware of phone number length they can give it a shot?

navyasrim: I tested with India and united states..it worked as expected...looks good to me...

navyasrim (approved):
looks good to me..

PR #616, opened 2022-12-22
blusquare sort by date implemented

navyasrim:
Bug : (Navya):: Blue squares on a person’s profile should always be in order by date. Right now they are in the order added. If I add one with a past date, it should show up in the correct order, NOT at the end of the line.

Fix : Implemented sort function,now  bluesquares are sorting by date.

### DIFF
--- a/src/components/UserProfile/BlueSquares/BlueSquare.jsx
+++ b/src/components/UserProfile/BlueSquares/BlueSquare.jsx
@@ -6,7 +6,8 @@ const BlueSquare = ({ blueSquares, handleBlueSquare, role, roles, userPermission
   return (
     <div className="blueSquareContainer">
       <div className="blueSquares">
-        {blueSquares ? (blueSquares.map((blueSquare, index) => (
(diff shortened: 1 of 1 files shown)


src/components/UserProfile/BlueSquares/BlueSquare.jsx
```diff
@@ -6,7 +6,8 @@ const BlueSquare = ({ blueSquares, handleBlueSquare, role, roles, userPermission
   return (
     <div className="blueSquareContainer">
       <div className="blueSquares">
-        {blueSquares ? (blueSquares.map((blueSquare, index) => (
+        {blueSquares ? (blueSquares.sort((a,b)=> a.date > b.date ? 1 : -1)
```
LeeSingChanYau, line 9: Looks Good! Tested it out locally and it works. The most recent blue recent blue squares show up in the right and the least recent at the left.

LeeSingChanYau (approved):
Looks Good! Tested it out locally and it works. The most recent blue recent blue squares show up in the right and the least recent at the left.

RJBornToWin (approved):
I approve the changes. Sorts bluesquares as oldest first.

PR #628, opened 2022-12-30
invert color for progress bar of task

juneluo:
### Description
Update the color of the progress bar (for tasks) as the request below:
requirement

### Mainly changes explained:
Introduce an optional parameter` invert `for function `getProgressColor`, so that when called for the task percentage, it shows the color inverted.
example

### How to test:
check into current branch
do npm install and ... to run this PR locally
go to dashboard→ Tasks

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -76,3 +76,13 @@ body,
 .bg--dark-sea-green {
   background-color: #c9d6c6;
 }
+
+.bg-bright-red {
+  background-color: #FF3300 !important;
+  /*  bright-red */
+}
+
+.bg-almost-red {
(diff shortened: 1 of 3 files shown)


one-community:
Summary of changes that I understand should be in this PR (more than what's in the description): the middle column mouseover text to say the right thing, that column to show that new data, the colors to be fixed for the bars to the right of task, and the Leaderboard bar sizes to be proportionate from 40 hours and below. That should wrap this all up!

Let me know if any of these are coming in a future PR instead

RJBornToWin (approved):
Approved.

PR #634, opened 2023-01-02
Shaun user management resized

ShaunSully88:
BUG: Admin → Other Links → User Management 
Page does not resize properly. Page should look good and be usable on tablet devices and phones. Use hiding info and symbols to make this work. 

FIX: Added Table component, included a noWrap class so that the data and cells remain same size regardless of User Device. This solution makes the table scrollable horizontally, and readable,  on all devices, all cells remain same size and nothing is actually resized based on device. I am going to make this my next project, restructuring the table based on User device.


rauleffting (approved):
Very simple and clean solution with nowrap. It's working quite well as you can see in the video below:

https://user-images.githubusercontent.com/29555732/210597706-dafe47bd-6d83-4b7e-966a-7491851bbc3d.mov

EvianTan (approved):

EvianTan:
The below images are after and before for management page on local and DEV.
image

PR #718, opened 2023-03-17
Lucas user hours set to zero

Lucas-E:
This PR is relative to task (PRIORITY LOW) Jae: Special Situation - User Hours Set to Zero

<hr/>

### Main changes

When a user's role is changed to Mentor or its hours are set to zero, the user will not be shown in the Leaderboard again. It only affects the Leaderboard, so the user's email still appears in the weekly reports.

<hr/>

### Related PRs (if any)

Please checkout to PR [#291(BE)](https://github.com/OneCommunityGlobal/HGNRest/pull/291) for the changes work correctly.

### How to test it

This frontend works fully only with the backend provided in the Slack conversation.

- Check the current branch in both the frontend and backend.
- Run npm install.
- Run npm run start:local on the frontend and npm run start on the backend.
- Login as an administrator.
- Create a test user.
- Check if the user is shown in the Leaderboard (it should be shown).
- Change the user's hours to zero and save.
- Check again (it should not be shown).
- Change the user's hours back to normal, save, and check again (it should be shown).
- Change the user's role to Mentor, save, and check again (it should not be shown).
- Change the user's role to any other than Mentor and check again (it should be shown).

<hr/>

### Notes

Please check if the user email is shown on the weekly reports page in each case.


src/components/UserProfile/VolunteeringTimeTab/VolunteeringTimeTab.jsx
```diff
@@ -88,10 +88,10 @@ const WeeklyCommittedHours = props => {
   return (
     <Input
       type="number"
-      name="weeklycommittedHours"
-      id="weeklycommittedHours"
-      data-testid="weeklycommittedHours"
-      value={props.userProfile.weeklycommittedHours}
```
vitorsaa2k, line 94: Please, make it back to `weeklycommittedHours` 
as you can see it's not showing the correct value:
image

PR #337, opened 2021-12-04
Dev new max personal record badge

navyasrim:
New Max Personal Record Badge Needs to be Updated
Files modified :
src/components/Badge/Badge.css
Added lines 41 to 55 for Max personal badge css style and postion
src/components/Badge/BadgeImage.jsx
Line 18 added condition to display Css name based on badge type
src/components/UserProfile/Badge.css
Added lines 49 to 62 for Max personal badge css style and postion
src/components/UserProfile/BadgeImage.jsx
Line 18 added condition to display Css name based on badge type

I made .CSS file changes for displaying the badge number of hours position as per the requirement
Also implemented the logic to display style based on badge condition like personal Max badge or other badge in css display.
As per Jae requirement I also adjusted the sizing of the badge number so i just changed the pixel ,font in other css method.
Badge_SizeAdjustment

PR #799, opened 2023-05-02
Ayush - Update env file to store the default password

ayush0904:
# Description
The current implementation of our codebase has a significant security vulnerability, as passwords are hardcoded in multiple locations throughout the code. This means that anyone with access to the repository can easily access these sensitive credentials. To address this issue, we need to implement a more secure approach for storing these passwords. One potential solution is to utilize a separate .env file to store all password information. By doing so, we can ensure that the passwords are kept confidential and are not visible to unauthorized individuals. This approach can significantly reduce the risk of security breaches and protect our system from potential attacks. To implement this solution, we will need to refactor the existing codebase to remove all hardcoded passwords and replace them with calls to the .env file. We will also need to ensure that the .env file is securely stored and that access to it is limited only to authorized individuals. Additionally, we may need to update any associated documentation or processes to reflect this change in our security approach.

## Related PRS (if any):
Related PR for backend #337
…

## Mainly changes explained:
Updated the .env file to inclure the password. (DEF_PWD=123Welcome!)
Made the changes in the places where the password is hardcoded.
…

## How to test:
check into current branch
do npm install and ... to run this PR locally
Check if the login is successgul from different users.

### DIFF
--- a/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
+++ b/src/components/UserProfile/AddNewUserProfile/UserProfileAdd.jsx
@@ -539,7 +539,6 @@ class AddUserProfile extends Component {
     } = that.state.userProfile;
 
     const userData = {
-      //password: '123Welcome!',
       password: process.env.REACT_APP_DEF_PWD,
       role: role,
       firstName: firstName,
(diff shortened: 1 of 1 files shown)


EvianTan (approved):

rauleffting:
Reviewed yesterday to help Yiyun figure out what was happening.

PR #847, opened 2023-05-23
Edwin - Removed option for Admins to reset password for Owners

EdwinLauM:
# Description

Screen Shot 2023-05-22 at 7 42 14 PM

## Mainly changes explained:
    1. Created a function to check permissions on owner user accounts to avoid duplicate code. 
    2. Added a conditional to check if the current user role has the correct permissions on owner user accounts. If it does, render the "reset password" button, if not don't. 
 
## How to test:
    1. check into current branch
    2. do `npm run start:local` to run this PR locally
    4. Login an admin account
    5. go to Other Links -> User Management
    6. Check that there is no "reset password" button for all owner accounts. 

### Before:

Admin Account: 
Screen Shot 2023-05-22 at 7 49 12 PM

### After:

Admin Account:
Screen Shot 2023-05-22 at 7 47 40 PM

Owner Account: 
Screen Shot 2023-05-23 at 10 49 28 AM

### DIFF
--- a/src/components/UserManagement/UserTableData.jsx
+++ b/src/components/UserManagement/UserTableData.jsx
@@ -29,7 +29,7 @@ const UserTableData = React.memo(props => {
       props.role,
       'addDeleteEditOwners',
       props.roles,
-      props.userPermissions) ? true : false
+      props.userPermissions)
     )}
      
   return (
@@ -96,6 +96,7 @@ const UserTableData = React.memo(props => {
           : ''}
       </td>
       <td>{props.user.endDate ? props.user.endDate.toLocaleString().split('T')[0] : 'N/A'}</td>
+      {checkPermissionsOnOwner() ? null : (
       <td>
         <span className="usermanagement-actions-cell">
           {checkPermissionsOnOwner() ? null : (
@@ -111,11 +112,10 @@ const UserTableData = React.memo(props => {
           )}
         </span>
         <span className="usermanagement-actions-cell">
-        {checkPermissionsOnOwner() ? null : (
-          <ResetPasswordButton user={props.user} isSmallButton />)
-          } 
+          <ResetPasswordButton user={props.user} isSmallButton />
         </span>
       </td>
+      )}
     </tr>
   );
 });
--- a/src/components/UserManagement/UserTableData.jsx
+++ b/src/components/UserManagement/UserTableData.jsx
@@ -99,7 +99,6 @@ const UserTableData = React.memo(props => {
       {checkPermissionsOnOwner() ? null : (
       <td>
         <span className="usermanagement-actions-cell">
-          {checkPermissionsOnOwner() ? null : (
             <button
               type="button"
               className="btn btn-outline-danger btn-sm"
@@ -109,7 +108,6 @@ const UserTableData = React.memo(props => {
             >
               {DELETE}
             </button>
-          )}
         </span>
         <span className="usermanagement-actions-cell">
           <ResetPasswordButton user={props.user} isSmallButton />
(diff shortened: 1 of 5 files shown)


lchua2314:
I do not have an owner account, so I will ask: Did you test an owner account? I don't know if devs are allowed owner accounts and I don't know how to make one.

src/components/UserManagement/UserTableData.jsx
```diff
@@ -24,6 +24,14 @@ const UserTableData = React.memo(props => {
     onReset(false);
   }, [props.isActive, props.resetLoading]);
 
+  const checkPermissionsOnOwner = () => {
+    return (props.user.role === 'Owner' && !hasPermission(
+      props.role,
+      'addDeleteEditOwners',
+      props.roles,
+      props.userPermissions) ? true : false
```
lchua2314, line 32: Just call "hasPermission" function and remove the ternary operators. hasPersmission() will return the correct boolean. Keep the "!" operator though.

Cirakuze:
Confirming the expected behavior logged in as an admin:

| BEFORE | AFTER |
|--|--|
| #847 BEFORE | #847 AFTER |

lchua2314 (changes requested):
Make the small change. Everything else looks good. Tested the feature and got the desired output:

Before:
Before

After:
After

Cirakuze:
It seems only the button inside the cell is hidden, and it leaves a a column of empty cells. Have you explored hiding the entire column, rather than just the buttons inside the cells?

eduhorta (approved):
Working as intended. Nice job! 👍
image

NidaZaki (approved):
The changes look good. Good job at hiding the entire cell for the "Owner".
Before:
PR#847 - BEFORE

After:
PR#847

lchua2314 (dismissed):
Nice job on the changes!

20chen-7 (approved):
The function is intended.
Screenshot 2023-05-23 at 11 40 55 AM
Screenshot 2023-05-23 at 11 40 22 AM

EdwinLauM:
> @EdwinLauM I do not have an owner account, so I will ask: Did you test an owner account? I don't know if devs are allowed owner accounts and I don't know how to make one. It looks like it'll work though.

I have tested it in an owner account. Edited the description to include a screenshot of how it looks in an owner user account

20chen-7:
> > @EdwinLauM I do not have an owner account, so I will ask: Did you test an owner account? I don't know if devs are allowed owner accounts and I don't know how to make one. It looks like it'll work though.
> 
> I have tested it in an owner account. Edited the description to include a screenshot of how it looks in an owner user account

susannagy:
Confirming that when I logged in as Admin user, in your branch that the column is not visible:
UI-manage-users

susannagy (dismissed):
👍

one-community:
Column should only be invisible IF filtering by "Owner", otherwise the password functions will be lost for other classes... but there will always only be one (maybe two) Owners, so hiding the column for that user class isn't something I care about.

src/components/UserManagement/UserTableHeader.jsx
```diff
@@ -47,7 +47,12 @@ const UserTableHeader = React.memo(props => {
       <th scope="col" id="usermanagement_resume_date">
         End Date
       </th>
-      <th scope="col" id="usermanagement_delete"></th>
+      {((
+        props.authRole !=="Owner" && 
+        props.roleSearchText !=="Owner") || 
+        props.authRole =="Owner") && (
```
Cirakuze, line 53: Seems this is the same condition as in `UserTableSearchHeader` lines from 56 to 58.
How about moving this into a helper function that'll be imported into both files and receives the thee values as params?
`(authRole, roleSearchText, authRole) => ((authRole !== "Owner" && roleSearchText !== "Owner") || authRole === "Owner");`

EdwinLauM: Moved the same condition from`UserTableHeader` and `UserTableSearchHeader` into a helper function and imported in both files.

PR #762, opened 2023-04-08
Jianjun and RJ - Performance Optimization

juneluo:
### Description
Complete the priority medium task of LEADERBOARD COMPONENT
Improve loading speed of other people’s Time Log: When I click the time bar in the leaderboard for someone, the loading speed to see their time log is really slow. I’d like to speed that up if possible. 
Remaining major open problems proposed by RJ are described in the analysis [here](https://www.dropbox.com/sh/e43le4bts99dxuw/AADHoHLWQVaf1FDHdqRgU2gNa?dl=0) 
To sum up, major points to check:
1. avoid unnecessary re-render
2. check the dependency of the hooks
3. convert the Timelog component from class to function

### Mainly changes explained:
1. set the value of prop `isDashboard` as true when the timelog component is called at the main page, remove a duplicated condition.
2. Introduce the state `summaryBarData` and redesign the passing process of the data for the SummaryBar component so that it won’t use the ones from the leader board.
3. Convert the Timelog component from class to function and use the hooks and states to control the re-render.

### How to test:
To test if the component runs normally:
Log in --> see dashboard --> click timelog page ==> shows the correct output
log out and log in as another account --> see dashboard ===>shows the correct output
add time entry ===>shows the correct output
To test the performance:
add console.log in the code and check console output to see how many times each part/function runs when open the page.

PR #855, opened 2023-05-27
fix timelog permission problem after improving performance

juneluo:
### Description
This is an updated version of a previous merged and reverted  PR #762 : fix the bug that when an admin view other user's dashboard, they do not get the permission to add time entry for that user (the orange add button doesn't show).

### Mainly changes explained:
Please check the #762  first to get the idea of what most of the code changes were for.
base on that, a replacement of the first prop for the `hasPermission` function (`auth.user`-->`auth.user.role`) will fix the bug, because the latter one is the correct datatype that is expected.

### How to test (this bug fix):
Log in as an admin-->
on the leaderboard, click the dot before a user's name to open their dashboard-->
see the add time entry button on that user's dashboard-->
add a time entry for them and save successfully.
img

### DIFF
--- a/src/components/Dashboard/Dashboard.jsx
+++ b/src/components/Dashboard/Dashboard.jsx
@@ -13,7 +13,7 @@ import { getTimeZoneAPIKey } from '../../actions/timezoneAPIActions';
 
 export const Dashboard = props => {
   const [popup, setPopup] = useState(false);
-  const [leaderData, setLeaderData] = useState(null);
+  const [summaryBarData, setSummaryBarData] = useState(null);
   const [userProfile, setUserProfile] = useState(undefined);
   let userId = props.match.params.userId ? props.match.params.userId : props.auth.user.userid;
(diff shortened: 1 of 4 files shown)


yufeiStella (approved):
It works as intended! Great work! The administrator is able to add time for a specific user successfully. Upon reloading the page, the added time is visible on the leaderboard.
We can see the "Add Time Entry for XXX":
Screen Shot 2023-05-27 at 14 20 32
After reloading the page:
Screen Shot 2023-05-27 at 14 21 28

AuroraLovesWhales (commented):
Hi @juneluo. I have tested your PR and it works as intended. The added time entry is immediately reflected on Leaderboard and the top bar. Great work! 

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/44281916/6662212b-1bf8-4db4-98c2-35944cf8c4d2

PR #858, opened 2023-05-27
only admin can edit bio in weekly summaries report

juneluo:
### Description
We expect only the admin/owner can edit the bio announcement posted status in the weekly summary report page. Other roles who can go to this page can see the status but cannot change it (now anyone who can see the summary report page can change it). 

### Mainly changes explained:
1.Pass the authorized role as the props to the component `FormattdReport` from `weeklySummariesReport`.
2.Check the role once and render different content (can edit the bio announcement or not. The screenshot below shows if a user can view but not edit it.)
img

### How to test:
Log in as admin/owner-->weekly summary report page-->click the toggle switch of any user to change their bio status
Log in as other roles with access to weekly summary report page (for example, manager)-->only show the label but no switch.

### DIFF
--- a/src/components/WeeklySummariesReport/FormattedReport.jsx
+++ b/src/components/WeeklySummariesReport/FormattedReport.jsx
@@ -154,68 +154,39 @@ const FormattedReport = ({ summaries, weekIndex, role }) => {
     }
   };
 
-  //These two similar but different(toggle switch onclick action) functions below are to wrap 
-  //up the actions that makes the content for each user's summary. Depending on whether the user
-  //has the permission, one of the two will be called in the map function.
-  //Checking the role once should be better than checking inside the map loop.
-  const summaryCardCanChange = (summary, index) => {
-    const hoursLogged = (summary.totalSeconds[weekIndex] || 0) / 3600;
-    const googleDocLink = getGoogleDocLink(summary);
-    const [isBioPosted, setIsBioPosted] = useState(summary.bioPosted);
+  const bioSwitch = (userId, bioPosted) => {
+    const [isBioPosted, setIsBioPosted] = useState(bioPosted);
     return (
-      <div
-        style={{ padding: '20px 0', marginTop: '5px', borderBottom: '1px solid #DEE2E6' }}
-        key={'summary-' + index}
-      >
-        <div>
-          <b>Name: </b>
-          <Link to={`/userProfile/${summary._id}`} title="View Profile">
-            {summary.firstName} {summary.lastName}
-          </Link>
-
-          <span onClick={() => handleGoogleDocClick(googleDocLink)}>
-            <img className="google-doc-icon" src={google_doc_icon} alt="google_doc" />
-          </span>
+      <div>
+        <div className="bio-toggle">
+          <b>Bio announcement:</b>
         </div>
-        <div>
-          {' '}
-          <b>Media URL:</b> {getMediaUrlLink(summary)}
+        <div className="bio-toggle">
+          <ToggleSwitch
+            switchType="bio"
+            state={isBioPosted ? false : true}
+            handleUserProfile={() => {
+              handleChangeBioPosted(userId, isBioPosted);
+              setIsBioPosted(!isBioPosted);
+            }}
+          />
         </div>
-        <div>
-          <div className="bio-toggle">
-            <b>Bio announcement:</b>
-          </div>
-          <div className="bio-toggle">
-            <ToggleSwitch
-              switchType="bio"
-              state={isBioPosted ? false : true}
-              handleUserProfile={() => {
-                handleChangeBioPosted(summary._id, isBioPosted);
-                setIsBioPosted(!isBioPosted);
-              }}
-            />
-          </div>
-        </div>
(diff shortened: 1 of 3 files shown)


suigunfjc (dismissed):
Tested, work as expected. Approved!
image
image

yufeiStella (approved):
It works! Only admin can change bio, for example, manager can't change a user's bio.
Screen Shot 2023-05-27 at 20 43 25
Screen Shot 2023-05-27 at 20 46 46

wang9hu (commented):
Hi @juneluo, I tested your code and it does work as intended, however, I am not sure what is the point of making a toggle button for a user that can't use it (e.g. manager), why not just making it like a label? 
Also, I compare the two similar functions and wonder if they could be combined since there is not too much difference. 
At last, the new added role props was not included in the type setting function (propTypes) in both files.

one-community:
> I am not sure what is the point of making a toggle button for a user that can't use it (e.g. manager), why not just making it like a label?

Outside of Owner/Admin, there will be people who can see this page but not be able to use the toggle and others who will be granted editing ability. We could make it a label for those without editing ability. Would coding that be difficult for users or classes granted special editing ability via the Other Links --> Permissions Management option?

one-community:
> In order to make one main feature in one PR, I will accomplish that permission management in another PR and move this PR progress forward first.

Yes, totally agree with this. Thank you!

src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
```diff
@@ -26,6 +27,7 @@ export class WeeklySummariesReport extends Component {
       error: this.props.error,
       loading: this.props.loading,
       summaries: this.props.summaries,
+      role: this.props.authRole,
```
Cirakuze, line 30: Does the `state.auth.user.role` data depend on `getWeeklySummariesReport` to finish before it can be accessed?
Looks like `summaries` is intended to be accessed only after `getWeeklySummariesReport` has finished. But is that also true for `role`?
I'm wondering whether `role` can be accessed directly from `props` without getting set to the state and have the same intended behavior.

src/components/WeeklySummariesReport/FormattedReport.jsx
```diff
@@ -153,61 +154,84 @@ const FormattedReport = ({ summaries, weekIndex }) => {
...
+        <b>Bio announcement:</b>
+        {bioPosted ? ' Posted' : ' Requested'}
+      </div>
+    );
+  };
+
+  const summaryCard = (summary, index, bioFunction) => {
+    const hoursLogged = (summary.totalSeconds[weekIndex] || 0) / 3600;
+    const googleDocLink = getGoogleDocLink(summary);
+    return (
```
Cirakuze, line 190: Maybe the `bioFunction` can be defined here between the lines 189 and 190 rather than having it be be a prop.

src/components/WeeklySummariesReport/FormattedReport.jsx
```diff
@@ -153,61 +154,84 @@ const FormattedReport = ({ summaries, weekIndex }) => {
...
-          </div>
-        );
-      })}
+      {bioCanEdit
+        ? alphabetize(summaries).map((summary, index) => {
+            return summaryCard(summary, index, bioSwitch);
+          })
+        : alphabetize(summaries).map((summary, index) => {
+            return summaryCard(summary, index, bioLabel);
+          })}
```
Cirakuze, line 234: If you define `bioFunction` within `summaryCard`, this can be: `{alphabetize(summaries).map((summary, index) => summaryCard(summary, index))}`

src/components/WeeklySummariesReport/FormattedReport.jsx
```diff
@@ -153,61 +154,84 @@ const FormattedReport = ({ summaries, weekIndex }) => {
...
+  const bioLabel = (userId, bioPosted) => {
+    return (
+      <div>
+        <b>Bio announcement:</b>
+        {bioPosted ? ' Posted' : ' Requested'}
+      </div>
+    );
+  };
+
+  const summaryCard = (summary, index, bioFunction) => {
```
Cirakuze, line 187: Maybe `bioFunction` can be defined here between `bioLabel` and `summaryCard` rather than `summaryCard` receiving it as a prop.

stevenpaalz (dismissed):
Confirming that changes worked as intended for volunteer, manager, and admin accounts. Note that volunteers can still access this report page even though it is not available in their menu.

Manager/Volunteer view:
Screenshot 2023-05-30 at 2 57 30 PM

src/components/WeeklySummariesReport/FormattedReport.jsx
Cirakuze: 1) Line 186: `const bioFunction = bioCanEdit ? bioSwitch : bioLabel;`.
2) Line 187: `const summaryCard = (summary, index) => {`.
3) Lines from 228 to 234: `{alphabetize(summaries).map((summary, index) => summaryCard(summary, index))}`.

Cirakuze (approved):

PR #891, opened 2023-06-07
Nida - WeeklySummaryReportPage- Refreshing the Weekly Summaries page stay on the same summary tab 

NidaZaki:
# Description
image

## Main changes explained:
- Setting `activeTab` to the latest selected tab by using `localStorage` for saving and getting the `tabSelection` value when the component is re-rendered.
- Setting `activeTab` to "Last Week" as a default value.

## How to test:
1. check into current branch
2. do `npm run start:local` to run this PR locally
3. login as admin user
4. Reports -> Weekly Summaries Report
5. Verify that the default tab selected is "Last Week", and when any other tab is selected, refreshing the page stays on the same selected tab. Also, when the page is revisited, the default tab selected should be "Last Week".

### After

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/32549172/3600c369-7eb8-4be7-8512-0f52fea9ed87

### DIFF
--- a/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
+++ b/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
@@ -17,24 +17,28 @@ export class WeeklySummariesReport extends Component {
     error: null,
     loading: true,
     summaries: [],
-    activeTab: '2'
+    activeTab: '2',
   };
 
   async componentDidMount() {
     await this.props.getWeeklySummariesReport();
-    window.addEventListener("popstate", () => {
-      localStorage.removeItem('tabSelection')
+    window.addEventListener('popstate', () => {
+      localStorage.removeItem('tabSelection');
     });
     this.setState({
       error: this.props.error,
       loading: this.props.loading,
       summaries: this.props.summaries,
-      activeTab: localStorage.getItem('tabSelection') === null ? '2' : localStorage.getItem('tabSelection'),
+      activeTab:
+        localStorage.getItem('tabSelection') === null ? '2' : localStorage.getItem('tabSelection'),
     });
   }
 
   componentWillUnmount() {
     localStorage.removeItem('tabSelection');
+    window.removeEventListener('popstate', () => {
+      localStorage.removeItem('tabSelection');
+    });
   }
 
   getWeekDates = weekIndex => ({
--- a/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
+++ b/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
@@ -17,16 +17,25 @@ export class WeeklySummariesReport extends Component {
     error: null,
     loading: true,
     summaries: [],
-    activeTab: '1',
+    activeTab: '2'
   };
 
   async componentDidMount() {
     await this.props.getWeeklySummariesReport();
+    window.addEventListener("popstate", () => {
+      localStorage.removeItem('tabSelection')
+    });
     this.setState({
       error: this.props.error,
       loading: this.props.loading,
       summaries: this.props.summaries,
+      activeTab: localStorage.getItem('tabSelection') === null ? '2' : localStorage.getItem('tabSelection'),
     });
+  }
+
+  componentWillUnmount() {
+    localStorage.removeItem('tabSelection');
+    window.removeEventListener("popstate");
   }
 
   getWeekDates = weekIndex => ({
@@ -46,6 +55,7 @@ export class WeeklySummariesReport extends Component {
     const activeTab = this.state.activeTab;
     if (activeTab !== tab) {
       this.setState({ activeTab: tab });
+      localStorage.setItem('tabSelection', tab);
     }
   };
 
--- a/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
+++ b/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
@@ -22,7 +22,6 @@ export class WeeklySummariesReport extends Component {
 
   async componentDidMount() {
     await this.props.getWeeklySummariesReport();
-    window.addEventListener('popstate', removeLocalStorageItem);
     this.setState({
       error: this.props.error,
       loading: this.props.loading,
@@ -34,7 +33,6 @@ export class WeeklySummariesReport extends Component {
 
   componentWillUnmount() {
     localStorage.removeItem('tabSelection');
-    window.removeEventListener('popstate', removeLocalStorageItem);
   }
 
   getWeekDates = weekIndex => ({
@@ -219,10 +217,6 @@ export class WeeklySummariesReport extends Component {
     );
   }
 }
-
-const removeLocalStorageItem = () => {
-  localStorage.removeItem('tabSelection');
-};
 
 WeeklySummariesReport.propTypes = {
   error: PropTypes.any,
--- a/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
+++ b/src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
@@ -27,12 +27,14 @@ export class WeeklySummariesReport extends Component {
       loading: this.props.loading,
       summaries: this.props.summaries,
       activeTab:
-        localStorage.getItem('tabSelection') === null ? '2' : localStorage.getItem('tabSelection'),
+        sessionStorage.getItem('tabSelection') === null
+          ? '2'
+          : sessionStorage.getItem('tabSelection'),
     });
   }
 
   componentWillUnmount() {
-    localStorage.removeItem('tabSelection');
+    sessionStorage.removeItem('tabSelection');
   }
 
   getWeekDates = weekIndex => ({
@@ -52,7 +54,7 @@ export class WeeklySummariesReport extends Component {
     const activeTab = this.state.activeTab;
     if (activeTab !== tab) {
       this.setState({ activeTab: tab });
-      localStorage.setItem('tabSelection', tab);
+      sessionStorage.setItem('tabSelection', tab);
     }
   };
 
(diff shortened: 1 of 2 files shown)


AnishPandita:
Hi @NidaZaki I tested the PR and can confirm that it displays the desired behaviour. The default tab is last week and when we change to different tab its stays on that after refresh. When we revisit the page it again reverts to last week default tab.

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/111321945/4e6704f2-ca85-4018-9683-e8dc114561cb

AnishPandita (dismissed):
Code is well structured and clean. Great Job.

OAkinbode (approved):
Hello. I think it works fine. 

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/104801652/dd72583e-25ca-40c9-b16c-5c978ab29ac4

And the code looks good to.

20chen-7 (approved):
The code

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/102696148/a140158a-2058-4f9c-b1de-486488240e23

 looks good, and the default is the last week. Switch to another week, the website does not refresh. Nice work!

bienzguoa (approved):
Works as intended! Code is clear!


https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/92126559/e3b0fa63-4d80-47e1-8664-148447454de9

eduhorta:
Hi, @NidaZaki!

I'm unable to test the PR because the timer has been reverted to the old model.

image

NidaZaki:
> Hi, @NidaZaki!
> 
> I'm unable to test the PR because the timer has been reverted to the old model.
> 
> image

Hi, Did you try doing `npm install` after pulling down this branch? If not, try doing `npm install` both in the FE and BE.

src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
```diff
@@ -17,18 +17,27 @@ export class WeeklySummariesReport extends Component {
     error: null,
     loading: true,
     summaries: [],
-    activeTab: '1',
+    activeTab: '2'
   };
 
   async componentDidMount() {
     await this.props.getWeeklySummariesReport();
+    window.addEventListener("popstate", () => {
```
leonzh2k, line 25: Could you explain the reasoning behind using this event listener? My guess is that it's supposed to cleanup localStorage after navigating to other pages (The state wouldn't be needed on those other pages) but could be wrong.

src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
```diff
@@ -35,7 +35,6 @@ export class WeeklySummariesReport extends Component {
 
   componentWillUnmount() {
     localStorage.removeItem('tabSelection');
-    window.removeEventListener("popstate");
```
leonzh2k: What's your reasoning behind removing the removal of this event listener? As it stands the event listener is never cleaned up before unmount, so it's leaking memory.

leonzh2k (changes requested):
@NidaZaki Functionality seems to be working as intended. 

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/49804462/9ac1f1e6-3327-4afa-b1f8-6e937719ad66

There are some questions I have about the code. Also I am curious if it's specified what the behavior should be when navigating to other pages, then back to this page.

src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
NidaZaki: Attaching a link to pop state event listener from mdn.

https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event

This specifically tracks browser events such as pressing the back button, or forward. Even gives you history (if you need it). So we “listen” to events where the user would press the back button on the browser. 

Yes correct, we don’t need it on other pages. And this PR should be ensuring clearing local storage.

src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
NidaZaki: Great catch! Let me double check if I accidentally removed it. Yeah I do think It should be present.

AriaYu927:
Hi @NidaZaki , I've tested your PR and it works as expected. I'll approve it after you address Leon's suggestions. Thank you!

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/94303659/9912c876-cb72-4110-936b-c42eca82ad25

yufeiStella:
It works perfectly!

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/111897624/11ecffe3-4afa-487f-b441-d5527ba104f7

eduhorta (approved):
Hi, @NidaZaki!
I guess something was missing and `npm install` did solve the problem. Thanks for the suggestion!

About the PR. The feature is working as intended, nice job! 👍

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/105796041/d92d3fd4-2fc5-4fce-9235-193de54a8dbc

leonzh2k (changes requested):
I don't think the event listener is being cleaned up properly. From testing, it seems that the event listener doesn't actually get cleaned up on component unmount. See the attached video and pay attention to the console (increase playback speed to see the point I'm making faster):

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/49804462/005818e0-ecff-40ca-9705-b0c0056159a0

The reason for this behavior is that the callback you passed into `addEventListener` does not reference the same function you passed into `removeEventListener`. To cleanup the listener properly in `componentWillUnmount`, the callback function passed into `addEventListener` and `removeEventListener` must reference the exact same function. 

See https://www.pluralsight.com/guides/how-to-cleanup-event-listeners-react (the relevant section is "Manual Event Listener Cleanup")

src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
leonzh2k: It seems that the `componentWillUnmount` runs when pressing the forward/back button, so I'm wondering if it's even necessary to use an event listener at all. What are your thoughts?

NidaZaki:
> I don't think the event listener is being cleaned up properly. From testing, it seems that the event listener doesn't actually get cleaned up on component unmount. See the attached video and pay attention to the console (increase playback speed to see the point I'm making faster):
> 
>  2023-06-12.15-38-11.mp4 
> The reason for this behavior is that the callback you passed into `addEventListener` does not reference the same function you passed into `removeEventListener`. To cleanup the listener properly in `componentWillUnmount`, the callback function passed into `addEventListener` and `removeEventListener` must reference the exact same function.
> 
> See https://www.pluralsight.com/guides/how-to-cleanup-event-listeners-react (the relevant section is "Manual Event Listener Cleanup")

@leonzh2k I took your suggestion and pushed a commit to address the issue. Check it out and let me know. 
Great catch again, and thanks for the resource.

NidaZaki:
> @NidaZaki I hate to suggest another change, but `componentWillUnmount` runs when pressing the forward/back button, so I'm wondering if it's even necessary to use an event listener. What are your thoughts?

Good call. Yeah I've tested it and `componentWillUnmount` does run on pressing back button. I wasn't aware it would trigger on pressing the back button but thanks for bringing that up! 

Pushed a commit that reflects the change.

src/components/WeeklySummariesReport/WeeklySummariesReport.jsx
NidaZaki: Pushed a commit that reflects the change. Replied to your comment below as I didn't see this comment on this thread earlier.

leonzh2k:
@NidaZaki One thing I just realized is that closing the tab/browser window does not run `componentWillUnmount`. Assuming the intended functionality is to have the default tab on "Last Week" when navigating to the Weekly Summaries page, here's how the current implementation doesn't satisfy this constraint:
1. When on the Weekly Summaries page, click on any tab other than "Last Week"
2. Close the tab/browser window
3. Open up the app again and go to the weekly summaries page
4. The tab will be on the last tab you were on, not "Last Week" (because `componentWillUnmount` didn't run and localStorage wasn't cleaned up)

To solve this problem I think the perfect solution is to use `sessionStorage` instead (everything in it is cleaned up when the tab/browser window is closed). Just replace every instance of `localStorage` with `sessionStorage` and verify that everything still works

NidaZaki:
> @NidaZaki One thing I just realized is that closing the tab/browser window does not run `componentWillUnmount`. If the intended behavior is to have the default tab always on "Last Week" when revisiting the Weekly Summaries page, here's how the current implementation doesn't satisfy this constraint:
> 
> 1. When on the Weekly Summaries page, click on any tab other than "Last Week"
> 2. Close the tab/browser window
> 3. Open up the app again and go to the weekly summaries page
> 4. The tab will be on the last tab you were on, not "Last Week" (because `componentWillUnmount` didn't run and localStorage wasn't cleaned up)
> 
> To solve this problem I think the perfect solution is to use `sessionStorage` instead (everything in it is cleaned up when the tab/browser window is closed). Just replace every instance of `localStorage` with `sessionStorage` and verify that everything still works

Interesting. Yeah, I tried this suggestion and changed `localStorage` to `sessionStorage` and that seems to have fixed the above issue. Thanks for the suggestion.
Pushed changes.

leonzh2k (approved):

papiasharmin (approved):
Everything is working as intended. Great work

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/86232710/17a2ae0c-18ce-4520-996f-30e606a36ab6

EdwinLauM (approved):
Works as intended! Good job! 

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/42288987/2e3f98d3-60ff-405f-8463-040f6fcf6480

one-community:
Great job everyone, I'll merge this as soon as the pending reviewers approve it and/or Nida says it is good to go.

PR #1076, opened 2023-07-31
set the default pageNum to 0 and prevent NaN start value

Cirakuze:
# Description
The Tasks list in the Dashboard was crashing.
Debugging showed the crush was caused by a `NaN` passed to `slice` within the `loadFunc` function.
The `NaN` was happening because the `loadFunc` function is not always called with an argument.
The `pageNum` param is used used to calculate an argument for a `slice`.
Setting the `pageNum` param of the function `loadFunc` to `0` fixes the crash.

## Main changes explained:
Set the `pageNum` param's default to `0` and prevent the `start` value from being `NaN`.

## How to test:
1. In the development branch, log in as a volunteer.
2. The Tasks in the Dashboard crashes.
3. Check out this branch and log in as the same user.
4. The Tasks should load.

| |
|--------|
| **BEFORE** TASKS - BEFORE |
| **The culprit** TASKS - arguments - code |
| **The output** TASKS - arguments - output |
| **AFTER** TASKS - AFTER |

### DIFF
--- a/src/components/TeamMemberTasks/TeamMemberTasks.jsx
+++ b/src/components/TeamMemberTasks/TeamMemberTasks.jsx
@@ -284,7 +284,7 @@ const TeamMemberTasks = React.memo(props => {
     }
   };
 
-  const loadFunc = useCallback(pageNum => {
+  const loadFunc = useCallback((pageNum = 0) => {
     if (teamList.length <= displayData.length) {
       setHasMore(false);
       return;


AaronPersaud (approved):
Hi Bada, this fix has resolved the crash on Volunteer/Manage/Mentor accounts

<b>Volunteer Account Team Member Tasks</b>

crash_fix_volunteer

- page doesn't crash anymore
- team member tasks render

<b>Admin Account Team Member Tasks</b> 

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/10265513/7c4c80b2-11ac-4fc9-935d-983222ac5aa3

- behaviour is the same as before

<b>User with many tasks</b>

image

Something to note: when there's a user with a very large amount of tasks, the page takes a lot longer to continue the infinite scroll and sometimes freezes. This is outside of this PR, and won't be noticed in main as no single user has this many tasks. We might need to do more optimizations and memoization in the future on this.

Excellent job! This was a much needed fix.

PR #1602, opened 2023-11-29
XIaoW_Add audio to forced paused timer

wang9hu:
### Description
Add audio to forcedPause when timer is forced paused or lost connection.

The purpose of this PR is to let users to know when their timer is paused unintentionally due to reasons like internet issues, especially when the app page is running in the background. 
This is not intended as a reminder when the user loads the page for the first time or upon refresh, because audio cannot be played directly upon opening a site or refresh.

### How to test
It is tricky to test this because if everything is running locally then disconnecting internet wouldn't help that much in this case. The tricky part is to keep the server running when break the connection locally, and the only way I can think of is to save on the frontend code to trigger a pseudo-refresh of the app, which isn't technically a browser level refresh but it breaks and reconnect the connection to server.


ptrpengdev:
Hi @wang9hu , could you provide more details on how to test the PR with forced paused and lost connection? I assumed that stopping the backend server would lead to a lost connection. However, it seems like not the case. The timer keep running even with a stopped backend server. Thank you!

image

image

Alforoan (changes requested):
I tested different ways of triggering the audio and it seems to work the first or the second time, but if you keep refreshing, it stops triggering the audio.


https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/118868270/8fcaa072-3ee2-4ee8-81a9-9995a1bbb0b2

wang9hu:
Thanks for the reviews, @ptrpengdev and @Alforoan! 

I have updated the description and hopefully it can help resolve some of the confusions. 

Let me know if you have more questions.

Alforoan (dismissed):
Everything works as it should. Nice!

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/118868270/0dd86d6f-beb8-4fe0-80bd-16e328fa2eab

PR #667, opened 2023-01-31
open all task folders in one click

juneluo:
### Description
Fixes bug priority low 7 of PROJECTS/TASKS/WBS COMPONENT:
Make clicking “All” on a WBS open all folders to show all tasks

### Mainly changes explained:
Load all tasks of the WBS when open the WBS task page but don't show the sub tasks in the initial state. Open all folders when the filterstate is set to 'all' by clicking the 'all' button.
https://user-images.githubusercontent.com/9314962/215760671-45b37427-eb5e-4f02-a904-dcc90c51e969.mov

### How to test:
check into current branch
do npm install and ... to run this PR locally
Logged in as Amin → Other Links → Projects → AAA USE THIS PROJECT → Click WBS Icon → Choose a WBS → Click “all”
img

PR #1748, opened 2023-12-27
Vishala add material

Vishala09:
# Description
Add Materials Page
Or Implements # (WBS) 3.3.7 (Phase 2)

## Related PR:
This frontend PR is related to the [#667](https://github.com/OneCommunityGlobal/HGNRest/pull/667) backend PR.

## Main changes explained:
- Created Add Materials Component
- Display of all materials
- Two ways to add units : a. Choose measurement from the list ,  b. Create new measurement 
- When measurement is newly added , some similarity checks are done.

## How to test:
1. check into current branch
2. do  `npm run start:local` to run this PR locally
3. Go to http://localhost:3000/bmdashboard/materials/add
4. Test for a presentable UI 
5. Page displays the list of materials to give users an overview before inserting the new material. Test for the materials display.
6. Test for 2 types of measurements insertion. Either select from the list or insert a new measurement.
7. Inserting a new measurement displays the similar measurements that are already available in the db if the similarity percentage is greater than 50%
8. Inserting a new measurement inserts the unit in units collection and displays it in the list of measurement units on the Add Materials Page
9. Editing one of the field (between new/old measurement) clears the other one restricting the user to add the measurement in only one way
10. Test for client side validations on the fields

[Untitled_ Dec 22,2023 9_17 PM.webm](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/65274029/32354ae4-7e12-463b-9354-afc73b5a6fd8)


tdkent (changes requested):
The basics work great! Adding a new material type adds a new doc to the correct db collection, and to the list of available types on the page.

Change requests:
- "Name": We need to run a check against existing types with the same name. Duplicates should not be allowed. The similarity check function would be better used here (see below).
- "New Measurement": I thought we were going to have a JSON file/db collection of units of measurement that is automatically available to users? I'm not sure if we need a means for the user to create new units, or if this is the right location for it.
- "Available Building Material Types": instead of a list of all the docs in the db, can we just show a set, with no duplicates, names only?
- Update `var` to either `const` or `let` as needed

Vishala09:
> The basics work great! Adding a new material type adds a new doc to the correct db collection, and to the list of available types on the page.
> 
> Change requests:
> 
> * "Name": We need to run a check against existing types with the same name. Duplicates should not be allowed. The similarity check function would be better used here (see below).
> * "New Measurement": I thought we were going to have a JSON file/db collection of units of measurement that is automatically available to users? I'm not sure if we need a means for the user to create new units, or if this is the right location for it.
> * "Available Building Material Types": instead of a list of all the docs in the db, can we just show a set, with no duplicates, names only?
> * Update `var` to either `const` or `let` as needed

Hi @tdkent 
I have worked on the mentioned changes. I have removed the display of static Types list. I have integrated the code base with JSON file for units. Duplicate materials cannot be added anymore.
https://www.loom.com/share/7ed2219eb0314b72ab982aa4748a06ed

tdkent (approved):

PR #2091, opened 2024-03-27
Nahiyan_Fixed white screen error on dashboard and timelog

Nahiyan-16:
# Description
Screenshot 2024-03-27 132813

## Related PRS (if any):
This frontend PR is related to the dev backend.

## Main changes explained:
-  Fixed task status not updating to `Started`
-  When deleting a project from the Projects page, it also deletes the WBS and tasks associated with it
- Added checks for existing tasks and WBS that do not have any projects associated with them but rather just linked with users

## How to test this error in **development**:
1. Follow these steps if you don't see the white screen error
2. check into the development branch
3. do `npm install` and `npm run start:local` to run this PR locally
4. Clear site data/cache
5. log as owner user
6. go to dashboard→ Other links → User Management → Create New User
7. Create a new owner account
8. Log onto the new owner account
9.  go to dashboard→ Projects → Create new project → added your new account to it
10. Go to your user profile to verify that you have added yourself to the new project
11. Go back to Projects, and delete the newly created project.
12. Now if you attempt to log time, go to timelog page, or go to dashboard page, you should see the white screen error

## To test my changes:
1. check into this branch
2. do `npm install` and `npm run start:local` to run this PR locally
3. Clear site data/cache
4. log as owner user
5. Repeat steps  6-11 from above
6. You should not see a white screen error when attempting to log time or going to either timelog page or dashboard. 

Here's how to replicate this error:

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/71025942/2a89327f-6615-4a35-8c76-f65759f39923

## Note:
Make sure to delete any unused accounts.
Also any unused projects

### DIFF
--- a/src/components/Projects/Projects.jsx
+++ b/src/components/Projects/Projects.jsx
@@ -15,7 +15,9 @@ import {
   deleteProject,
   modifyProject,
 } from '../../actions/projects';
+import { deleteTask, fetchAllTasks } from '../../actions/task';
 import { getPopupById } from '../../actions/popupEditorAction';
+import { deleteWbs, fetchAllWBS } from '../../actions/wbs';
 import Overview from './Overview';
 import AddProject from './AddProject';
 import ProjectTableHeader from './ProjectTableHeader';
@@ -120,6 +122,16 @@ handleSort = (e)=>{
     let { projectId } = this.state.projectTarget;
     // request delete on db
     this.props.deleteProject(projectId);
+    // request delete wbs and tasks of the project on db
+    this.props.fetchAllWBS(projectId);
+
+    this.props.state.wbs.WBSItems.forEach(wbs => {
+      this.props.fetchAllTasks(wbs._id)
+      this.props.state.tasks.fetchedData[0]?.forEach(task => {
+        this.props.deleteTask(task._id)
+      })
+      this.props.deleteWbs(wbs._id)
+    })
     // disable modal
     this.setState({ showModalDelete: false });
   };
@@ -363,6 +375,10 @@ export default connect(mapStateToProps, {
   postNewProject,
   deleteProject,
   modifyProject,
+  deleteTask,
+  fetchAllTasks,
   getPopupById,
+  deleteWbs,
+  fetchAllWBS,
   hasPermission,
 })(Projects);
--- a/src/components/Projects/Projects.jsx
+++ b/src/components/Projects/Projects.jsx
@@ -15,9 +15,7 @@ import {
   deleteProject,
   modifyProject,
 } from '../../actions/projects';
-import { deleteTask, fetchAllTasks } from '../../actions/task';
 import { getPopupById } from '../../actions/popupEditorAction';
-import { deleteWbs, fetchAllWBS } from '../../actions/wbs';
 import Overview from './Overview';
 import AddProject from './AddProject';
 import ProjectTableHeader from './ProjectTableHeader';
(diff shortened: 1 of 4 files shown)


Nahiyan-16:
@wang9hu @hetvi11110

anandsesha:
Hi, I have checked into the said branch in [#820](https://github.com/OneCommunityGlobal/HGNRest/pull/820) backend PR and the development branch frontend. 

After starting builds locally and clearing site data/cache,

- Dashboard is not visible on login, show the following console errors
dashboard error 1
dashboard error 2

- I navigated to http://localhost:3000/usermanagement  and tried creating a new owner account and it showed the below error.
new user error
 
- The timelog page error is as shown below
Timelog

iven-yao (dismissed):
I just reviewed this pull request, and everything looks perfect.
- first I create a temp project and add an owner account to it:
image
- after deleting the project, user could still view dashboard, timelog and update time entries without any problem.
image
image

mohabbasd (dismissed):
Hello, I tested your PR and it works as expected. Below is a video for your reference.

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/91215370/2ff7ef22-758b-4315-af95-d83a58b8f815

wang9hu:
Hi @Nahiyan-16, I have reviewed your PR and I understand what you are trying to solve here, but I think there are some other scenarios that still needs to cover in order to fix this thoroughly.  

For example, just like I mentioned in the bug doc, one case is that when a user adds time entry to a task/project and then get removed from that task/project (not trying to delete task or project, but to remove the user from task resources), the logged time entry will have a taskId/projectId that doesn't match the userId (because the backend will use userId to look for the task/projects, and it won't return the task/project if the users is removed), therefore causing a mismatch problem. 

And also, I think another issue is, what the time entry should be like if its task/project got deleted? Time entry is connected to a lot of things, one example is, in userprofile, there are some fields like tangiblehours and intangblehours, which are increased upon logging time entries, so do we need to remove the time of those time entries from them as well? I think we need to consider it thoroughly before making any changes.

Sarthak32 (dismissed):
Hello, I tested your PR it works as expected !
Screenshot 2024-03-30 150342
Screenshot 2024-03-30 152852
Screenshot 2024-03-30 153037

one-community (changes requested):
> Hi @Nahiyan-16, I have reviewed your PR and I understand what you are trying to solve here, but I think there are some other scenarios that still needs to cover in order to fix this thoroughly.
> 
> For example, just like I mentioned in the bug doc, one case is that when a user adds time entry to a task/project and then get removed from that task/project (not trying to delete task or project, but to remove the user from task resources), the logged time entry will have a taskId/projectId that doesn't match the userId (because the backend will use userId to look for the task/projects, and it won't return the task/project if the users is removed), therefore causing a mismatch problem.
> 
> And also, I think another issue is, what the time entry should be like if its task/project got deleted? Time entry is connected to a lot of things, one example is, in userprofile, there are some fields like tangiblehours and intangblehours, which are increased upon logging time entries, so do we need to remove the time of those time entries from them as well? I think we need to consider it thoroughly before making any changes.

Good points! Would the easiest solution here be to just save the data when deleting projects? 

I think the original reason we didn't do that was just trying to keep our data use down while doing all this testing. Now that Phase 1 and this related testing is almost done, we could just store it rather than "deleting it forever". Would that be any problem?

PR #2153, opened 2024-04-12
XiaoW_Hotfix of Timelog Page not displaying

wang9hu:
# Description
add authId as the default displayUserId if no id present from url param


abdel-lall (changes requested):
Hi @wang9hu , you made a typo here: it should be `authUser.userid` instead of `authUser.userId`.

wang9hu:
> Hi @wang9hu , you made a typo here: it should be `authUser.userid` instead of `authUser.userId`.

Thanks, fixed, nice catch!

PR #2278, opened 2024-05-22
Nahiyan_Hotfix for Header UI issues

Nahiyan-16:
# Description
Fixed time entry form header text in light mode
Reverted timer size in main header
Fixed main logo not displaying on smaller viewports

### DIFF
--- a/src/components/Header/Header.css
+++ b/src/components/Header/Header.css
@@ -78,9 +78,3 @@
     display: none;
   }
 }
-
-@media screen and (min-width: 1400px) and (max-width: 1700px) {
-  .owner-message {
-    display: none;
-  }
-}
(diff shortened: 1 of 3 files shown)


KyreneAF (dismissed):
- Message disappears if screen size less than 1400px allowing header to compress.
- Timer modal stays the same color in light mode and dark mode and functions correctly.
- Timer font-size adjusts dynamically.

Screenshot 2024-05-22 at 1 20 46 PM
Screenshot 2024-05-22 at 1 20 24 PM

KyreneAF (approved):
- Message disappears if screen size less than 1400px allowing header to compress.
- Timer modal stays the same color in light mode and dark mode and functions correctly.
- Timer font-size adjusts dynamically.

PR #2449, opened 2024-07-12
XiaoW_Hotfix of Timer disconnection reminder 

wang9hu:
# Description

Because the client websocket can not be closed when internet is cut off (suddenly lost server connection), the readyState will be stuck OPEN, so here we need to use a custom readyState to mimic the real readyState, and when internet is cut off, the custom readyState will be set to CLOSED, and the user will be notified to refresh the page to reconnect to the server.


JatinAgrawal94 (changes requested):
Can you please mention more specifically what's the bug and how to test it.

PR #2197, opened 2024-04-26
Peterson_impl_improve_user_Management_page_search_function_to_include…

peterson337:
# Description
This pull request was crafted with the intention of enhancing the input fields responsible for filtering the users displayed in the table on the user management page. Previously, when typing a user's name, it would appear in the table. However, if there was an extra space at the end of the name, the user would disappear from the table. This issue has been addressed in this pull request.

## Related PRS (if any):
None
…

## Main changes explained:
The user management component has been modified to improve the input responsible for filtering users.
…

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. Clear site data/cache
4. Log in as an administrator or owner
5. go to other links→ User Management
6. Please enter something in all input fields for first name, last name, search, and email, with spaces added at the beginning or end.
7.  Please enter numbers in input fields for Wkly Committed Hrs or search, with spaces added at the beginning or end

## Screenshots or videos of changes:
Before my improvement: 
a

After my improvement:
Improve User Management page search function to include spaces

## Note:
None

PR #2819, opened 2024-10-26
Fixed PR#2197

pallavithorat:
# Description

This pull request resolves an issue in the User Management search functionality, where adding a space after a search term (e.g., "Li ") was intended to limit the search to exact matches for that term, rather than partial matches. Previously, the system returned all names containing the search term, even if an exact match with a trailing space was entered (e.g., entering "test " would incorrectly return names like "testuser" instead of just "test").

# Fix Details
Modified the search function in UserManagement.jsx to handle trailing spaces correctly.
Now, when a user enters a search term with a space at the end, the search is limited to names that exactly match the given term, rather than including partial matches.
This ensures that entering "Li " will only return names that are exactly "Li", providing more precise search results as requested.

Fixes #2197 
Task

## Main changes explained:

Updated the search functionality in User Management to support exact-match searches when a trailing space is added to the FirstName/ LastName. Now, entering "test " will return only users with the exact name "test", instead of including partial matches like "testuser". This improves search accuracy as requested.

## How to test:
1. check into current branch
2. do npm start to run this PR locally
3. Clear site data/cache
4. log as admin user
5. go to dashboard→ Other Links -> User Management

6. Verify Functionality behaves as expected with the new exact-match functionality in First Name and Last Name Columns.  Please see the attached screenshot for confirmation

## Screenshots of changes:
# Before Pictures:
Before 1
Before 2

# After Pictures:
After 1
After2

### DIFF
--- a/src/components/UserManagement/UserManagement.jsx
+++ b/src/components/UserManagement/UserManagement.jsx
@@ -286,54 +286,52 @@ class UserManagement extends React.PureComponent {
       userTableItems: this.userTableElements(userProfiles, rolesPermissions, timeOffRequests, darkMode, editUser)
     })
   }
-
+  
   filteredUserList = userProfiles => {
     return userProfiles.filter(user => {
-      // Applying the search filters before creating each table data element
-      const search = result => {
-        if (typeof result === 'string') {
-          return result
-            .toLowerCase()
-            .trim()
-            .replace(/\s+/g, '');
-        }
-        const numberToString = String(result)
-          .trim()
-          .replace(/\s+/g, '');
-        const stringToNumber = Number(numberToString);
-        return stringToNumber;
-      };
-
+      
+      const firstNameSearch = this.state.firstNameSearchText || '';
+      const lastNameSearch = this.state.lastNameSearchText || '';
+  
+     
+      const firstName = user.firstName.toLowerCase();
+      const lastName = user.lastName.toLowerCase();
+  
+      
+      const trimmedFirstNameSearch = firstNameSearch.trim().toLowerCase();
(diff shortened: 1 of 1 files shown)


samarth9201 (commented):
There seems to be some kind of bug here. When I tried to search in the user management table, it did not give me any results

Screenshot (170)
Screenshot (171)
Screenshot (172)

Muhideenthegreat (changes requested):
Bugs To Be Fixed:

- When I went straight from the dashboard in dark mode and tried to switch back to the light mode on the User Management page, the table format remained in dark mode.

Screenshot 2024-10-26 at 12 13 23 PM


- Also there seem to be a bug in the search algorithm (for the first and last name search bar) as it doesn't give suggestion until the entirety of the name is typed in which isn;t how a search function should work. 
Screenshot 2024-10-26 at 12 24 04 PM
Screenshot 2024-10-26 at 12 24 31 PM

Verification:

- The search function works well for the email and weekly hours commited search.

Screenshot 2024-10-26 at 12 24 55 PM

pallavithorat:
Okay I will look into it,
Thanks!

anji-annavarapu:
I tested the filter search, and it works when an exact name match is provided in the data. However, if I don’t know the full or exact name, it doesn’t return any results when I enter only the first two letters. I understand that it’s currently set up for exact matches, but this makes searching a bit difficult. My suggestion is to allow it to return names that start with the letters I enter.
Screenshot 2024-10-26 120543
Screenshot 2024-10-26 120606
Screenshot 2024-10-26 120640
Screenshot 2024-10-26 120704
Screenshot 2024-10-26 120732

PR #3206, opened 2025-02-27
Aureliano - improve link requirement for timelog submission

AurHubertMax:
# Description
image

## Related PRS (if any):
none

## Main changes explained:
- Updated file `src\components\Timelog\TimeEntryForm\TimeEntryForm.jsx` to include RegEx pattern
- Pattern should allow any link starting with `https://` and disable dropbox links if it is copied from the address bar instead of the share link.

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. Clear site data/cache
4. log as any user
6. go to dashboard→ Tasks→ task→stop timer and log time
7. Verify that any link starting with `https://` are allowed to be submitted with the exception of dropbox links which are copied from the address bar instead of the share link.
8. Verify that logging time still works.

https://github.com/user-attachments/assets/b5b9e499-78a3-4a3d-85a5-65c70e5c44a2


Manoj99Q (approved):
Verified localhost urls and non shared dropbox links are blocked.
Shared Dropbox links are accepted

https://github.com/user-attachments/assets/7b5bf1a0-a17f-4f45-a094-5412f0fd7851

CodewithKoushica (changes requested):
I tested different links and found that the following cases are currently allowed, but they should be blocked:

1) Localhost & Private Network URLs:
- https://localhost
- https://127.0.0.1:5000
- https://10.0.0.1/dashboard
These should be restricted to prevent logging links pointing to local or private network addresses.

2) Invalid Dropbox Links:
- https://www.dropbox.com/folder/file.pdf
This is not a valid Dropbox share link and should be blocked to enforce the correct sharing format.

Please let me know if these are working as expected.


https://github.com/user-attachments/assets/57339a1a-2c21-49d0-9d99-6d97cb8d9a82

AurHubertMax:
> I tested different links and found that the following cases are currently allowed, but they should be blocked:
> 
> 1. Localhost & Private Network URLs:
> 
> * https://localhost
> * https://127.0.0.1:5000
> * https://10.0.0.1/dashboard
>   These should be restricted to prevent logging links pointing to local or private network addresses.
> 
> 2. Invalid Dropbox Links:
> 
> * https://www.dropbox.com/folder/file.pdf
>   This is not a valid Dropbox share link and should be blocked to enforce the correct sharing format.
> 
> Please let me know if these are working as expected.
> 
>  Screen.Recording.2025-02-27.at.9.21.06.PM.mov

Thank you for catching this error @CodewithKoushica! I've updated the regex so it should block the proper links now, here's a video of it working:

https://github.com/user-attachments/assets/28358d19-b40c-40f5-8090-b476c54efda1

03kshitij:
Any links starting with https:// is working and also able to log the time

Screenshot 2025-03-04 at 6 53 47 PM
Screenshot 2025-03-04 at 6 54 25 PM

HHHHHeloise:
All links starting with https:// is working and also able to log the time and these are properly blocked as expected. Overall, this implementation works well.

CodewithKoushica (dismissed):
Thank you for working on that, everything works as expected now!
Screenshot 2025-03-07 at 3 30 31 PM

vikasreddynomula (dismissed):
Doesn't allow dropbox links that are not share links

Screenshot 2025-03-07 at 5 06 22 PM



Doesnt allow localhost links

Screenshot 2025-03-07 at 5 02 15 PM

Screenshot 2025-03-07 at 5 15 24 PM

vasavi-139 (dismissed):
The functionality works when:
--> I haven't submitted the link
--> When I have used an https link
--> When I have used my drop box
Screenshot 2025-03-07 190740

Screenshot 2025-03-07 191801
 link

SravyaKotra:
The update ensures that any https:// link works correctly, and time logging functions as expected without issues.

sidhu47h (changes requested):
The regex filter is not working for all the cases. It didn't allow me when i used the following cases
* No link
* http:localhost:3000

Allowed when a valid https with google or dropbox is present but its failing for the following case
https://dropbox:5000

this is not a valid url so we need to also ensure that no special characters except for `.` should be allowed after the dropbox or google. We need to ensure that the same thing followed by `com` and `.`

Screenshot 2025-03-18 at 4 55 45 PM
Screenshot 2025-03-18 at 4 56 15 PM
Screenshot 2025-03-18 at 4 56 41 PM

AurHubertMax:
> The regex filter is not working for all the cases. It didn't allow me when i used the following cases
> 
> * No link
> * http:localhost:3000
> 
> Allowed when a valid https with google or dropbox is present but its failing for the following case https://dropbox:5000
> 
> this is not a valid url so we need to also ensure that no special characters except for `.` should be allowed after the dropbox or google. We need to ensure that the same thing followed by `com` and `.`
> 

Hi @sidhu47h, a link is always required when logging time, and localhost links are NOT allowed, as stated in the description. Furthermore, shared dropbox links would have special characters other than `.` after the `dropbox` for their share keys so I can't filter out any special characters after the `dropbox`

myeeli (approved):
I have checked out the current PR and logged in as a volunteer user. I was able to log time for:  

https://docs.google.com/document/d/1p9XSzQuLxxmBHRMbCRxx4Xywz5645zuXzzhhqiig52g/edit?usp=sharing
https://www.google.com/

However, I was unable to log time for:  

http://localhost:3000/dashboard# 
https://www.dropbox.com/home/Manvitha%20Yeeli

This behavior is expected, so I am approving the PR.
PR - 3206 -2
PR 3206 -3
PR 3206 - 1

pdp5 (dismissed):
I have reviewed the changes, and everything is working as expected. The RegEx pattern correctly allows links starting with https:// while preventing Dropbox links copied from the address bar. Additionally, time logging functionality remains unaffected and works as expected.

Invalid Dropbox URL
with local URL
without description and url
without url

karnati007 (changes requested):
Tested the submission of links starting with "https://", verifying that they are accepted while Dropbox links copied from the address bar are correctly blocked. However, the description mentions that localhost addresses should be disallowed, but they are currently being accepted. Additionally, tested the stop timer and log time functionality, which is working as expected. Everything else is functioning correctly, but the localhost validation needs to be addressed.

AurelianoPR4
AurelianoPR3
AurelianoPR2
AurelianoPR

AurHubertMax:
> Tested the submission of links starting with "https://", verifying that they are accepted while Dropbox links copied from the address bar are correctly blocked. However, the description mentions that localhost addresses should be disallowed, but they are currently being accepted. Additionally, tested the stop timer and log time functionality, which is working as expected. Everything else is functioning correctly, but the localhost validation needs to be addressed.
> 

Hi @karnati007, thank you for reviewing the PR! Please double check if you've checked into the current branch, because I cannot reproduce the errors you listed, as shown below. Let me know if the time log still allows localhost connections after double-checking it. Thank you!

image

PR #3319, opened 2025-03-26
Anjali - Tab title to display user name

Maddila-Anjali:
# Description
This PR addresses a bug where the browser tab title fails to reflect the correct user's name when navigating between profiles from the User Management page. 

## Main changes explained:
- Implemented a useEffect hook to update the document.title whenever the userProfile object changes.
- Ensured the tab title reflects the full name (firstName + lastName) of the user being viewed.

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. Clear site data/cache
4. log as admin user
5. go to dashboard→ select a user from the list
6. verify that the persons name appears on the tab once opened on a new tab

https://github.com/user-attachments/assets/2cac5415-6302-4190-b38f-28d698173e97

### DIFF
--- a/src/components/UserProfile/UserProfile.jsx
+++ b/src/components/UserProfile/UserProfile.jsx
@@ -220,6 +220,12 @@ function UserProfile(props) {
     loadUserTasks();
   }, [props?.match?.params?.userId]);
 
+  useEffect(() => {
+    if (userProfile?.firstName && userProfile?.lastName) {
+      document.title = `${userProfile.firstName} ${userProfile.lastName}`;
+    }
+  }, [userProfile]);  
+
   const checkIsProjectsEqual = () => {
     const originalProjectProperties = [];
     originalProjects?.forEach(project => {
--- a/src/components/UserProfile/UserProfile.jsx
+++ b/src/components/UserProfile/UserProfile.jsx
@@ -221,10 +221,12 @@ function UserProfile(props) {
   }, [props?.match?.params?.userId]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


angad363 (dismissed):
I followed the instructions and verified that the user's name now correctly appears in the browser tab when navigating between profiles. The implementation works as expected.

Refer to the attached video for confirmation. Approving the PR.


https://github.com/user-attachments/assets/f2857e78-35b3-4a85-8d00-d7d6b85828bb

one-community (changes requested):
This isn't how I'd like it, it says “user” instead of the person’s name. It doesn’t update to the person’s name until clicked. It shouldn’t need a click, I want it to say the name when command clicked and BEFORE being visited so that I can open a bunch of tabs and know who they are BEFORE navigating to them.

vaibhavkoladiya (changes requested):
Thanks for the update! I noticed the user's name does show up, but there's a delay. Ideally, it should appear in the tab title right when the profile opens in a new tab.

That would make it easier to switch between profiles without confusion. Might be worth updating the logic to set the name right away.  Thank you.

https://github.com/user-attachments/assets/702d7b6a-33e6-4f4d-8c55-109fb08f9f10

Amalesh-A (approved):
I checked out your branch, reviewed your code, followed your instructions, and observed that the functionality works as expected.

Amalesh-A (commented):
I checked out your branch, reviewed your code, followed your instructions and observed that the functionality works when switching to the tab. It has to be noted that it only reflects the user's name when we switch to the tab else it says "User".

PR #5166, opened 2026-04-18
Chirag - Added Sorting Logic to drop-off & no-show rate columns in Participation table

ChiragBellara:
# Description
Screenshot 2026-04-17 at 8 15 40 PM

## Related PRS (if any):
This is **NOT** related to any other PRs.

## Main changes explained:
- Added logic to sort data based on no-show, drop-off and title values

## How to test:
1. check into current branch
2. do `npm install` and `npm run start:local` to run this PR locally
3. Clear site data/cache
4. log as admin user
5. navigate to `/communityportal/reports/participation `
6. verify that when you click on any of the headers in the Participation table, it sorts the data

https://github.com/user-attachments/assets/c3923307-848a-42f3-9fda-33c079f7fb59


Rajasrivatsansrinivasan (dismissed):
Reviewed PR #5166 by Chirag. Checked out the branch locally and tested the Participation table sorting on the community portal reports page. Verified that the No-show rate, Drop-off rate, and Title column headers now sort the table data correctly in both ascending and descending order when clicked. No issues were observed during testing. Approving.

Image
Image

kunchalasireesha (changes requested):
The sorting logic is currently using lexicographical (string) comparison, which causes 'Event 100' to appear before 'Event 11'. We should implement a natural sort so that the numeric values are respected

No show off rate and NO drop off rate is working as expected

ChiragBellara:
> The sorting logic is currently using lexicographical (string) comparison, which causes 'Event 100' to appear before 'Event 11'. We should implement a natural sort so that the numeric values are respected
> 
> No show off rate and NO drop off rate is working as expected

The lexicographical ordering is as expected. In a real world scenario we will not have events named "Event 1" and "Event 11". Even if that happens, lexicographical sorting or no sorting at all is the best way to go about showing the events initially. 
However, the scope of this PR is limited to sorting the No show off rate and No drop off rate only. Those are the only changes made in this PR.

one-community:
Thank you all, merging!

PR #5373, opened 2026-07-07
Abhishek_Fix: Duplicate header rendering in Total Org Summary Dashboard

Abhi-R0211:
## Bug
image

The Total Org Summary dashboard rendered its entire top header (title, date-range filter, comparison dropdown, Save as PDF button) twice — two identical consecutive rows on the same page.

## Root Cause
The header and date-range modal had been refactored into reusable components (`ReportHeader`, `DateRangeModal`), but the original inline JSX for both was never removed from the component's return block. Both the inline version and the new component version were rendering simultaneously, wired to the same shared state.

## Fix
Removed the leftover inline `<Row>` header block and inline `<Modal>` date-picker block. `ReportHeader` and `DateRangeModal` are now the single source of truth for this UI.

## Testing
- Verified only one header renders, light and dark mode (screenshots below)
- Verified date-range dropdown (Previous Week / Current Week) still   updates state and label correctly
- Verified Save as PDF still works
- Verified single date-range modal (Select Date Range) still opens/works

## Note for the team
While testing, found an unrelated pre-existing bug: switching the comparison dropdown from "No Comparison" to any other option (Week/Month/Year Over Year) throws `TypeError: roleDistributionStats is not iterable` in RoleDistributionPieChart.jsx, crashing the page. Confirmed this also happens on unmodified `development` — not caused by or fixed in this PR. Flagging separately for triage.

Screenshots:
Screenshot 2026-07-06 195226
Screenshot 2026-07-06 195241
Screenshot 2026-07-06 202145


DeepighaJ (changes requested):
- Tested the PR locally, only 1 header appears in the total org summary, no duplicates.
- Filters Current week, previous week, selected date range works fine.
- On selecting date range loading takes longer than usual but loaded after a while.
- Pdf generation works successfully.
-As mentioned the No comparison filters throwing errors for any of the values filtered .
-**Issue**: Select date range popup needs dark mode improvement including calendar. Is this out of scope too?

Image
Image
Image
Image
Image
Image

HemanthNidamanuru (changes requested):
Hi Abhishek,

I tested this PR locally.

The header refactor, date-range dropdown, Save as PDF, and single date-range modal are working fine.

One issue I noticed is that the calendar component is not aligned properly in dark mode. Also, the calendar needs proper navigation controls so users can move between dates/months easily.


Image

Image

Image

Image

Image

Image

kzou55 (changes requested):
Hi Abhishek,

I ran and tested the PR locally. 

## Verified
- Only one header

  Image

  Image

- Date filter

  https://github.com/user-attachments/assets/f56bc45d-8445-4ea3-bbfc-9907507b6f9b

- Comparison Options crash as stated

  Image

- PDF button works

  Image


## Issues
1. Date Component Header has poor contrast/visibility in Dark Mode. 
2. Date picker does not have Dark Mode applied to it. 

  Image

one-community:
Thank you all, merging!  Requested Adhya create a new action item for all the other things you found, great job team!

PR #5376, opened 2026-07-09
Abhishek 🔥: Blue Square Stats donut chart label overlap, dark mode contrast, and border spikes

Abhi-R0211:
## Bug
The Blue Square Stats donut chart had three related issues:
1. Small-segment labels overlapped each other, with leader lines crossing awkwardly (e.g. "0 (0%)", "961 (7%)", "83 (1%)" clustered together at the top of the ring)
2. Segment number labels were hardcoded black, not adjusted for dark mode
3. Thin adjacent segments showed stray spiked lines past the outer ring edge

Screenshot 2026-07-09 000254
Screenshot 2026-07-09 000214

## Root Cause
- No handling for small slices in chartjs-plugin-datalabels, so all labels defaulted to center-anchored and collided when several thin slices sat next to each other
- Label color was hardcoded '#000000'
- Doughnut segment borderWidth defaulted to centered border alignment, causing strokes to spike outward at sharp angles between thin adjacent slices
- This card's height is equalized with its row sibling (Team Stats) via Bootstrap's flex-stretch grid behavior, so there's very little
  vertical slack to push small-slice labels outside the ring without getting clipped

## Fix
- Small slices (<10% of total) no longer render an in-chart label at all; instead, their value and percentage are shown directly in the legend text (e.g. "Missing Hours: 961 (7%)"), matching the existing pattern already used successfully in the Role Distribution chart elsewhere in this codebase. This avoids fighting the tight, height-constrained space inside the ring entirely.
- The large slice (currently "Missing Both Hours & Summary" at ~92%) keeps its in-chart label as before, now with per-segment,
  contrast-aware text color instead of hardcoded black, so it's legible in both light and dark mode regardless of segment color.
- Zero-value slices no longer render a redundant "0 (0%)" label.
- Added borderAlign: 'inner' so segment border strokes stay within the arc instead of spiking past the outer edge.
- Legend now stacks below the chart on narrow screens instead of wrapping awkwardly in a cramped side column.

## Testing
- Login → Reports → Total Org Summary → Blue Square Stats
- Run backend on development branch
- Verified in light mode, dark mode, desktop, and mobile (simulated + real narrow viewport)
- Verified the shared Volunteer Status donut chart (same component) still renders correctly
- Screenshots attached for all four states

Screenshot 2026-07-09 011546
Screenshot 2026-07-09 011604


DeepighaJ (changes requested):
- Tested the PR locally below are findings.
- Only the largest segment (92%) is labeled on the chart, while the smaller segments are no longer displayed.
- The connector lines for the smaller slices remain visible even though they no longer connect to labels.
- The very small slices (1% and 0%) are difficult to distinguish visually.

Image
Image

HemanthNidamanuru (changes requested):
Hi,

I tested this PR locally.

The Blue Square Stats chart is working well overall. Small slices no longer show labels inside the donut, values and percentages are displayed in the legend, the large slice label is readable, and the legend stacks properly in smaller screen width.

Dark mode also looks good.

One small issue I noticed is that there are still thin line spikes/label lines visible outside the donut near the small slices.

Image

Image

Abhi-R0211:
Hi @HemanthNidamanuru and @DeepighaJ , I have fixed the reported issues. Please re-verify.

Screenshot 2026-07-13 140439
Screenshot 2026-07-13 140428

HemanthNidamanuru (changes requested):
Hi Abhishek,

I tested this PR locally.

The previously reported issues have been resolved. The thin line spikes and label lines are no longer visible outside the donut, and the chart looks good in both light and dark modes.

However, I noticed a new issue. Although the small slices have been removed, the chart still shows a value of 16 even after changing the date range. This does not match the values shared in the PR. Could you please confirm whether this is the expected behavior?


Image

Image

Abhi-R0211:
Hi @HemanthNidamanuru — good catch, but I confirmed this is expected behavior, not a bug.
 
The Blue Square Stats count is scoped strictly to whatever date range is selected, filtered on each infringement's own date. The three ranges you tested aren't contradictory once you look at where the boundaries actually fall — they're disjoint (or nearly disjoint) slices of the same underlying data, not overlapping views that should agree:
 
- June 28, 2015 – June 28, 2026 → **16**
- June 28, 2024 – June 28, 2026 → **0**
- June 26, 2026 – July 16, 2026 → **1755**
Most of the test infringement data turns out to be dated very recently (within the last few weeks), with only a small amount of older legacy data going back to 2015, and apparently none at all in the 2024–June 28, 2026 window.
 
To confirm the numbers are genuinely additive and consistent rather than a real bug, I ran a combined range spanning all three windows:
 
- June 28, 2015 – July 16, 2026 → **1771**
That's exactly `16 + 1755 = 1771`, confirming the counts are accurate and correctly filtered by date — just unevenly distributed across time in this dataset. No code changes needed here.
 
Let me know if you'd like me to double check any other specific date ranges, but this one's working as intended.

HemanthNidamanuru (dismissed):
Hi Abhishek,

Thank you for the clarification.

I rechecked the PR, and the previously reported issues have been resolved. The thin line spikes and label lines are no longer visible outside the donut, and the chart looks good in both light and dark modes.

Based on your clarification regarding the displayed value, I am approving this PR.


Image

kzou55 (dismissed):
Hi Abhishek, 

I ran and tested the PR locally. 

The changes look good and no longer have the specified issues. 
  Image

  Image


  Image

  Image

one-community:
Thank you all, merging!

PR #4762, opened 2026-01-26
Sayali taking over for Shravan - Email Management UI Fixes (Completing Apoorva's PR #4492 and #4731)

Shravan-neelamsetty:
# Description
image

This PR completes the Email Management UI improvements by addressing all 11 critical issues identified in Chaitanya's comprehensive feedback on PR #4492. This work takes over from Apoorva's previous attempts and provides a complete, tested solution for the Email Announcement System frontend.

Fixes: High Priority Issues from Chaitanya's feedback (PDF attached to PR #4492)
Implements: Email Management System Enhancement - Complete UI/UX Fixes

### Frontend:

Supersedes #4492 (Apoorva - original with conflicts)
Supersedes #4731 (Apoorva - clean version)

### Backend:

This frontend PR is related to OneCommunityGlobal/HGNRest#1945 backend PR (Chaitanya - Mailchimp Replacement)
To test this frontend PR, you need to checkout the backend PR #1945 on branch [chaitanya-mailchimp-replacement](https://github.com/OneCommunityGlobal/HGNRest/pull/1945)

#### Issue 1: URL Navigation

- Updated src/components/Announcements/platforms/email/index.jsx - Added React Router integration with useHistory and useLocation hooks
- Updated src/routes.jsx - Added missing routes for /announcements/email/send and /announcements/email/outbox
- Created getViewFromPath() function to parse URLs and sync with component state
- Updated navigation handlers to push URL changes via history.push()

#### Issue 2: Search UX in Templates

- Updated EmailTemplateList.jsx - Removed conditional wrapper around controls section
- Updated Controls (search, sort, create button) to stay visible during loading with disabled state
- Updated Results count to only show when not loading

#### Issue 3: Variable Labels in Preview

- Created getVariableTypeIcon() helper function with emoji icons (📄, 🎥, 🖼️, etc.)
- Updated EmailTemplateEditor.jsx preview modal to show icon + variable name
- Updated EmailTemplateList.jsx info modal to match same icon format
- Deleted references to non-existent variable.label property

#### Issue 4: Remove Validate API

- Deleted validateEmailTemplate import and function calls from EmailTemplateEditor
- Deleted API call block from handleSave function
- Updated dependencies array and PropTypes to remove validate references

#### Issue 5: Template Creation Bug

- Created createdTemplateId state variable to track newly created templates
- Updated handleSave to capture and use template ID after creation
- Updated Save logic to use templateId || createdTemplateId for updates
- Updated Preview function to use same ID logic

[...]

PR #4726, opened 2026-01-17
Fixed 4 critical UI/UX issues in the Email Management system:

apoorvajainrp21:
## Summary
Fixed 4 critical UI/UX issues in the Email Management system: 

1. ✅ **VIDEO Badge Type Support** - Added VIDEO type with proper icon display in EmailTemplateEditor
2. ✅ **Clear Draft Functionality** - Completely resets all form states when clearing draft
3. ✅ **Offline Alert Auto-Dismiss** - Alert automatically dismisses when connection restored
4. ✅ **Preview Image Sizing** - Large images constrained with max-width in preview
5. ✅ **BONUS: Auto HTML Wrapping** - Videos automatically wrapped in clickable YouTube buttons

## Files Changed
- `IntegratedEmailSender.jsx` - Added handleClearDraft and offline alert auto-dismiss
- `IntegratedEmailSender.module.css` - Added image size constraints for preview
- `EmailTemplateEditor.jsx` - Added VIDEO type support with icon
- `utils.js` - Added automatic HTML wrapping for images/videos

### Test #1 - VIDEO Badge Type
1. Navigate to `/announcements/email/templates`
2. Create or edit a template
3. Add a new variable
4. Check "Type" dropdown - should see "Video" option
5. Select "Video" type
6. ✅ Verify VIDEO icon badge appears next to variable name

### Test #2 - Clear Draft Functionality
1. Navigate to `/announcements/email/sender` (or `/sendemail`)
2. Fill out form: subject, recipients, content, select template
3. Wait for auto-save (draft indicator appears)
4. Click "Clear Draft" button
5. Confirm warning popup
6. ✅ Verify ALL fields reset: subject, recipients, content, template, header/footer images, schedule options

### Test #3 - Offline Alert Auto-Dismiss
1. Open email sender page
2. Disconnect WiFi/internet
3. ✅ Verify red offline alert toast appears
4. Reconnect WiFi/internet
5. ✅ Verify alert automatically disappears (no manual close needed)

### Test #4 - Preview Image Sizing
1. In email sender, use a template with an image variable
2. Add a large image URL (e.g., `https://picsum.photos/2000/2000`)
3. Click "Preview & Send"
4. ✅ Verify image is constrained and doesn't overflow preview
5. ✅ Verify preview is scrollable if content is long

### Test #5 - Auto HTML Wrapping (BONUS)
1. Create/edit template with a VIDEO type variable
[...]

PR #4492, opened 2025-12-04
Sai Shravan Neelamsetty taking over for Apoorva - Redo of email management UI improvements PR#4670

apoorvajainrp21:
# Description
This PR implements comprehensive UI/UX improvements for the Email Management system, specifically focusing on the EmailOutbox component's modal layout and batch information display. The changes address critical usability issues where email information and batch details were being cut off or not displaying properly.

**Implements:** Email Management System Enhancement - EmailOutbox Modal & Batch Display Improvements

## Related PRS (if any):
NEW VERSION OF THIS IS https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/4731

This frontend PR is related to Chaitanya's backend PR for email management [PR1945](https://github.com/OneCommunityGlobal/HGNRest/pull/1945)
To test this frontend PR, you may need to checkout Chaitanya's backend PR if the `/api/email-outbox/:emailId` endpoint is not yet merged to the main backend.
…

## Main changes explained:

**EmailOutbox.jsx** - Complete Modal & Batch Display Overhaul

#### Modal Layout Fixes:
- **Updated Modal component**: Changed from `centered` prop to custom top-aligned positioning with `style={{ maxWidth: '95%', margin: '1.75rem auto' }}` to maximize visible space
- **Removed maxHeight restrictions** from ModalBody to prevent content truncation
- **Increased modal width** from default `xl` to 95% viewport width to accommodate all content without horizontal scrolling

#### Email Information Section Restructure:
- **Redesigned from single-column to 2-column grid layout** using Bootstrap's `row` and `col-md-6` classes
- **Added 6 new fields** that were previously missing:
  - Email ID with word-break styling to prevent overflow
  - Started timestamp (green text when available, "N/A" when null)
  - Completed timestamp (green text when available, "N/A" when null)
  - Total Batches count (blue bold)
  - Sent Recipients count (green bold)
  - Failed Recipients count (red bold)
- **Updated field labels** from "Created At"/"Updated At" to "Created"/"Last Updated" for consistency
- **Applied color coding**: Blue for IDs/counts, Green for success timestamps/counts, Red for failures

#### Batch Information Table Implementation:
- **Created complete batch table** with dark header (`table-dark`) and striped rows (`striped hover`)
- **Implemented 5-column structure**:
  1. **Email Batch column**: Shows batch number and recipient count with responsive mobile fallback
  2. **Status column**: Displays status badge with corresponding icon (checkmark, clock, warning)
[...]

PR #5464, opened 2026-08-22
Aditya - Fix primary 2 buttons on the Announcements email dashboard

AdiDubbs:
# Description

The Announcements email dashboard has two cards, **Email (Send, Outbox)** and **Manage Templates**, and they are the only route into the email module. Clicking either one returned the Not Found page.

`/announcements` was registered as an exact match in `routes.jsx`, so every path beneath it - `/announcements/email/send`, `/announcements/email/outbox`, `/announcements/email/templates`, and the platform tabs - matched no route at all and fell through to the catch-all `NotFoundPage` at the end of the Switch.

## Related PRS (if any):

Replaces #4726, now closed. No backend changes.

## Main changes explained:

1. Removed `exact` from the `/announcements` route so sub-paths match. It is the only route with that prefix, so nothing is shadowed. Added a comment explaining why it must not be exact, since it now reads as inconsistent with its neighbours.
2. Corrected twelve social platform icon paths in `Announcements/index.jsx` from relative to root-relative. They resolved only at the parent route, so once the sub-paths began working they returned 404s. Four icons already used a leading slash and were unaffected.
3. Added `src/__tests__/routes.test.jsx`, covering the four email sub-paths against the real route table. Verified that it fails if `exact` is reintroduced.
4. Added `src/components/Announcements/platforms/email/__tests__/index.test.jsx`, which clicks each of the two dashboard cards and asserts the resulting path and view.

## How to test:

**This page requires the `sendEmails` permission. It comes from the **Owner role**, not from a personal permission — which matters for step 3 below.**

1. Check out this branch, run `npm install`, then `npm run start:local`
2. Clear site data and cache
3. Log in as an Owner and let the dashboard finish loading. Don't type `/announcements` straight into the address bar right after logging in: `ProtectedRoute` checks permissions before the role list has loaded and redirects every permission-gated page to the dashboard (a pre-existing issue, not from this PR).
4. Open **Other Links → Send Emails** from the header menu (goes to `/announcements`) - both cards should be visible
5. Click **Email** - loads `/announcements/email/send` with the sender
6. Switch to the **Outbox** tab - URL becomes `/announcements/email/outbox`
7. Click **Back to Dashboard**, then **Manage Templates** - loads `/announcements/email/templates`
8. Paste `/announcements/email/templates` into the address bar and reload - it should load rather than 404
9. Confirm the social platform icons render on each of those pages
10. Repeat steps 4-9 in dark mode
11. Negative cases: `/announcementsfoo` should still show Not Found, and a signed-out visitor to `/announcements/email/send` should be redirected to login rather than the 404 page
12. Run `npm run test:changed`

If you are still redirected to the dashboard at step 4, wait for the dashboard to finish loading and use the menu link again. This is a pre-existing issue in `ProtectedRoute` affecting every route with `routePermissions`, being logged separately.

## Screenshots of changes:

| | |
|---|---|
| **Before — either card, or any email sub-path direct, returned Page Not Found** <br><br> Page Not Found at an announcements email sub-path | **After — the two dashboard cards, light mode** <br><br> Announcements email dashboard showing both cards, light mode |
| **The two dashboard cards, dark mode** <br><br> Announcements email dashboard showing both cards, dark mode | **Send Email — reachable for the first time** <br><br> Send Email page |
| **Outbox — rows populating, so the backend calls succeed** <br><br> Outbox page with rows | **Email Templates — also where the twelve social icons had been 404ing** <br><br> Email Templates page with social platform icons rendering |

## Note:

The four files #4726 added are already on development - that work was carried forward through #4492, #4731 and #4762. What remained outstanding was the routing fault, which this PR fixes.

Separate pre-existing issues found while testing, not addressed here, to be logged on the bugs doc:

- `/sendemail` crashes on direct navigation — `EmailSender.jsx` reads `location.state.state`, which is undefined unless the route is reached via a `history.push` carrying nested state.
- A few of the social platform logo PNGs have no alpha channel, so they render as opaque white rectangles against the dark theme.


iAbhi001 (changes requested):
---

Hi @AdiDubbs,

Thanks for working on this fix!

During local testing, navigating to `/announcements` (as well as the sub-paths like `/announcements/email/send`) immediately redirects back to the main dashboard instead of rendering the announcements cards/views.

Could you clarify:

1. **Required Permissions/Role Setup:** I noticed in the PR description that this requires the `sendEmails` permission. Is there a specific permission flag or seed data required locally aside from the Owner role to prevent the redirect to the dashboard?
2. **Permission Guard:** Is the route guard redirecting before permissions finish loading from the auth context/state?

Please let me know how to properly configure the local user/role so I can verify the sub-path routing and complete the review.

DeMoliT1on (approved):
Tested on local with `Owner`/`Admin` and `Member` accounts. The routes seems to be working properly now. All the sub-routes for announcement page can be viewed when the user have the permission. And redirects to dashboard when the user does not have the required permission.

Image

Image

AdiDubbs:
> Hi @AdiDubbs,
> 
> Thanks for working on this fix!
> 
> During local testing, navigating to `/announcements` (as well as the sub-paths like `/announcements/email/send`) immediately redirects back to the main dashboard instead of rendering the announcements cards/views.
> 
> Could you clarify:
> 
> 1. **Required Permissions/Role Setup:** I noticed in the PR description that this requires the `sendEmails` permission. Is there a specific permission flag or seed data required locally aside from the Owner role to prevent the redirect to the dashboard?
> 2. **Permission Guard:** Is the route guard redirecting before permissions finish loading from the auth context/state?
> 
> Please let me know how to properly configure the local user/role so I can verify the sub-path routing and complete the review.

Thanks for the careful review - and both good questions. @iAbhi001 . I'll answer your questions below.

1. No extra seed data is needed. The page needs sendEmails, which only Owner has by default on dev. But I don't think that is what redirected you.

2. Yes, you're right. ProtectedRoute builds the permission list from state.role.roles, and if that hasn't loaded it falls back to an empty array, so "not loaded yet" and "not allowed" look identical to it and it redirects. The role list is normally cached in the browser, but step 2 of testing steps says to clear site data, which wipes it, and nothing refetches it unless you visit User Management or Permissions Management. So you'd have been redirected on any account, including Owner.

To test: log in as Owner, open /permissions once so the roles load, then go to /announcements. I've updated the testing steps so this doesn't catch anyone else.

The guard behaviour is pre-existing and affects every route with routePermissions, not just this one, so I'll log it on the bugs doc rather than widen this PR. Thanks for working out what was actually happening.

iAbhi001:
Hi, @AdiDubbs, As soon as load the /announcements page, I am still being forwarded to dashboard. I think, I may be missing something. Please let me know your thoughts.

iAbhi001 (changes requested):
Hi, @AdiDubbs, As soon as load the /announcements page, I am still being forwarded to dashboard. I think, I may be missing something. Please let me know your thoughts.

DeepighaJ (approved):
I use Owner account only, but i see page not found when i try to access /permissions 

- Verified /announcements displays both announcement cards.
- Verified Email loads /announcements/email/send with the sender displayed.
- Verified Outbox tab changes the URL to /announcements/email/outbox and loads correctly.
- Verified Manage Templates loads /announcements/email/templates.
- Verified direct navigation to /announcements/email/templates loads successfully instead of showing a 404.
- Verified /announcementsfoo displays Not Found.
- Verified signed-out access to /announcements/email/send redirects to the Login page instead of 404.

Note:
- Some social platform icons doesn't show their respective color in dark mode. If this is existing issue and out of scope of this PR, then can be ignored. Approving since current functionality works fine.

Image

Image

Image

Image

Image

Image

Image

AdiDubbs:
@iAbhi001 Thanks for checking again. The redirect isn't from this PR: `ProtectedRoute` redirects before the role list has loaded, which happens on every permission-gated page. My step 3 was also wrong, since `/permissions` doesn't exist. I've updated the steps: after logging in, let the dashboard finish loading, then open **Other Links → Send Emails** instead of typing the URL.

AdiDubbs:
@DeepighaJ Thanks for the review, and good catch on the 404. `/permissions` isn't a real route (Permissions Management is `/permissionsmanagement`), so that step was wrong. I've replaced it in the instructions with opening Announcements from the menu once the dashboard has loaded. 

Regarding the icons, the social icon colors in dark mode are an existing issue and out of scope for this PR: it only fixes the icon image paths, not their colors.

one-community:
Thank you all, merging!

PR #5517, opened 2026-09-09
Linlin - fix: formatting of header on Medium and Small screens

linlin-husky:
# Description

Stabilises the `Header` layout across mobile, tablet and desktop, restores the
owner/dev-environment message that had gone missing from it, and stops the mobile
navigation panel from displacing the page content beneath it.

* **Motivation & Context**: Header elements (user info, navigation links, the centre
  logo) squashed, wrapped unpredictably or disappeared depending on the viewport. On
  narrow screens the header stacked into five sparse rows, the timer controls folded
  into a grid two or three rows tall, and opening the hamburger menu grew the header
  and pushed the dashboard down. Separately, the owner message component had been
  dropped from the header entirely, leaving only its orphaned styles behind.

## Related PRs (if any):

None — frontend only, **no backend change required**.

### Header structure and desktop layout

* `Header.jsx` — user info elements (avatar, welcome dropdown, notification bell) moved
  into a dedicated top-row container (`userInfoRow`), navigation links into a bottom-row
  container (`menuContainer`).
* `Header.module.css` — flex/grid structure reworked with explicit gaps and minimum
  constraints so the logo cannot be compressed and nav items keep consistent spacing.
* `public/header-test.png` — header logo asset.

### Restored the owner message

`OwnerMessage` renders the owner / dev-environment banner from `state.ownerMessage`. It
had been replaced in `.centerSection` by the logo image and its import removed, so the
banner vanished from the header; the leftover `.navbarOwnerMessage` styles and the mock
in `Header.test.jsx` were the only signs it had ever been there. It is restored
alongside the logo, with the layout driven entirely from CSS — the two stay plain
siblings in the markup.

### Mobile (`<= 768px`)

* A named grid puts the warning message and the hamburger on one row in adjacent
  columns, with the logo centred on its own row above. `.centerSection` and
  `.rightSection` are dissolved with `display: contents` so their children become grid
  items directly — **no markup change**.
* The expanded nav panel is absolutely positioned, so opening it overlays the content
  below instead of growing the header. It keeps `grid-area: nav`, which for an
  absolutely positioned grid child supplies the containing block and anchors it under
  the message row with no hand-maintained offsets. Its background matched the exact
  colour Bootstrap paints the Navbar, so it dissolved into the bar it hangs from — a
  lighter surface and a layered shadow now separate the two.
* The collapsed panel gets a full-width row of its own; left in the narrow menu column
  its 220px minimum would have reserved that width permanently and squeezed the message.
* Menu rows carry their own `padding-left`: Bootstrap's `.navbar-nav` declares
  `padding-left: 0` at equal specificity, so the container's own padding depended on
  stylesheet order.

### Tablet / small laptop (`768–1280px`)

Timer and logo share the first line rather than taking one each. Above **850px** the
logo is lifted out of flow and centred on the full header width; below that the timer's
right edge reaches the middle of the header, so the logo stays in flow on its own row
instead of overlapping the timer controls.

### Shared components

* `Timer.module.css` — below 460px `.btns` became a CSS grid, wrapping the five controls
  onto two or three rows. They fit on one line (readout 100px + five ~24px icons ≈ 240px
  once gaps and padding come off, comfortable at 320px), so the grid rules are replaced
  with tighter spacing.
* `OwnerMessage.module.css` — on mobile the banner stacks (icons above, text below) and
  both align right so the block reads as belonging with the menu it sits against.
  `column-reverse` achieves the ordering without touching the markup.

**Both files are shared by four headers** — `Header`, `BMHeader`, `KIHeader` and
`CPHeader` — so these two changes affect all of them. Worth a glance at the other three
at narrow widths.

### SonarQube clean-up

All 12 findings raised on this PR are resolved:

* **Duplicate selectors (`css:S4666`)** — six same-scope pairs merged (four in the base
  scope, one in `<= 768px`, two in `768–1280px`). Duplicates *across* media queries are
  ordinary redefinition and were left alone. Each merge keeps the declarations that
  already won the cascade, so computed styles are unchanged.
* **Commented-out code (`css:S125`)** — a stale `/* position: relative; */`, redundant
  because the live rule it merged into already declares it.
* **Unused `currentWidth` (`javascript:S1854`)** — the whole resize effect was dead: it
  debounced to reading `window.innerWidth` into a variable nothing consumed, so the
  effect goes rather than just the assignment.
* **Unused `CardText` import** in `Badge.jsx` — `b8e8505f6` replaced the `<CardText>`
  usage with a `<div>` and left the import behind.

### Desktop is untouched

Every layout rule added here lives inside a media query. The `@media (width > 1280px)`
block was diffed byte-for-byte against `development` after each change. The one
exception is a `flex-direction` + `flex-wrap` pair collapsed into `flex-flow: row wrap`,
which the repo's pre-commit stylelint gate required — computed styles are identical.

## How to test:

1. Check out this branch.
2. `npm install`, then **`npm run start:local`** to run locally. *(There is no `dev` script in this
   repo — `npm run dev` will fail.)*
3. Clear site data/cache.
4. **Start the backend.** The warning banner is rendered from `GET /ownerMessage`; with
   no backend running the component renders an empty container and the text simply will
   not appear.

5. Log in as an administrator or owner. Confirm the banner shows in place of the logo. If no custom message is set you should still see the message slot with its edit controls — this is what lets an admin create the first message.

5b. Log in as a volunteer. Confirm you see the logo, not the banner — even though a global standardMessage is set. Then have an admin set a custom message and confirm the volunteer's logo is replaced by it.

6. Resize across viewports and check each band:
   * **> 1280px** — unchanged from `development`.
   * **850–1280px** — timer and logo on one line, logo centred on the full header width.
   * **768–850px** — logo drops to its own row below the timer; confirm it never
     overlaps the timer controls.
   * **<= 768px** — timer / logo / (message + hamburger) on three tight rows.
7. On mobile, open the hamburger menu and confirm the panel **floats over** the dashboard
   below rather than pushing it down, that the header height does not change, that the
   panel is visually distinct from the bar it hangs from, and that the menu items are
   indented rather than flush to the left edge.
8. Confirm the timer controls stay on a single horizontal row down to ~320px.
9. Check the Badge page still renders correctly (the `CardText` → `div` swap and the
   removed import).
10. Verify in [dark mode](https://docs.google.com/document/d/11OXJfBBedK6vV-XvqWR8A9lOH0BsfnaHx01h1NZZXfI).

Before change: the layout is a bit out of order and the navigation bar items are cramped together

Heading Before change

After change

https://github.com/user-attachments/assets/dad5d9fc-f6ce-4cef-b8b3-98a82231f61d

## Note:

* **Unrelated changes carried on this branch.** `b8e8505f6` and `4b133d1ab` also touch
  files that are not part of this fix: `package.json` (adds a `preview` script),
  `UserState/{UserStateBadge,UserStateDisplay,UserStateModal}.jsx` and
  `TeamMemberTasks/FollowupCheckButton.jsx`. Flagging so reviewers do not read them as
  part of the header work — happy to split them out if preferred.
* **Asset name.** `public/header-test.png` is the production logo despite the `test` in
  its filename; worth renaming before merge.
* Full test suite green: **2223 passed, 14 skipped, 0 failed.** Production build passes.
  stylelint and ESLint report no errors on the changed files.


iAbhi001 (dismissed):
Hi @linlin-husky,

Tested the PR locally across multiple breakpoints:

* Verified responsive layout across mobile (~375px), tablet (~768px), around the 1128px medium breakpoint, and full desktop screens.
* The two-row structure (`userInfoRow` and `menuContainer`) keeps navigation links cleanly spaced without overlapping.
* The center logo remains consistently centered, uncompressed, and fully visible across all viewports.
* Tested both light and dark mode; text contrast and visual assets render properly.

**Note:** The PR currently has a merge conflict with `development` in `src/components/Header/Header.jsx`. Once that conflict is resolved and the branch is updated, this should be good to merge.

Approving! Thanks for the fix.

Image

Image

shubhamjakhete (changes requested):
Requesting changes: the responsive layout currently hides the entire `userInfoRow` on screens `<=768px` using `hideInMobile`. That row contains several important user actions, including the notification bell, profile/avatar, Welcome menu, Dark Mode, Update Password, and Logout.

In the current development version, mobile users still have separate mobile-specific controls for these actions. Those mobile equivalents are removed in this PR, so the new layout creates a functional regression where mobile users lose access to key header actions.

Please preserve mobile-accessible versions of these controls while keeping the new responsive header layout.

PR #5103, opened 2026-04-05
Veda permission question sets

Charitha2009:
# Description
Screenshot 2026-04-04 at 10 54 12 PM
Screenshot 2026-04-04 at 10 54 26 PM

## Related PRS (if any):
This frontend PR is related to the [#2237](https://github.com/OneCommunityGlobal/HGNRest/pull/2237) backend PR.

## Main changes explained:

- Added new permission create_form_question in user roles
- Enabled non owner users to create and manage question sets
- Built dynamic form creation UI with customizable questions
- Introduced reusable question sets not tied to a specific job
- Added fixed fields (name, email, etc.) with support for custom fields
- Added optional ad link support
- Integrated permission based API checks

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. Clear site data/cache
4. log as admin user
5. go to http://localhost:5173/jobformbuilder
6. verify: 
- Permission based access
- Create and edit question sets
- Add custom fields
- Reuse question sets
- Dark mode works

## Screenshots or videos of changes:
Screenshot 2026-04-04 at 10 55 07 PM
Screenshot 2026-04-04 at 10 55 23 PM

PR #3989, opened 2025-08-29
Dharmik/frontend/application/job posting page/function create permission to create different question sets

Dharmik1710:
# Description
image
image

## Related PRS (if any):
This backend PR is related to the [#1685](https://github.com/OneCommunityGlobal/HGNRest/pull/1685) frontend PR.
To test this backend PR you need to checkout the [#1685](https://github.com/OneCommunityGlobal/HGNRest/pull/1685) frontend PR.
…

## Main changes explained:
1. Permission-based question set creation - Different user roles can create different types of question sets
2. Question set management system - Create, edit, delete, and organize question sets by category
3. Enhanced job form builder - Integration with existing job posting system
4. Category-based organization - Question sets organized by Engineering, Marketing, Design, etc.
5. Search and filtering - Find question sets by category and search terms

## How to test:
Refer the below video for testing.

## Screenshots or videos of changes:
https://github.com/user-attachments/assets/4a282d74-b471-4b09-a9aa-edaed6a4347e

PR #5527, opened 2026-09-11
Gayatri - fix(job-form-builder): question numbers now follow position on Move Up/Down (Ticket #1)

sawantgayatri19:
# Description
question

Fixes # 1 (bug list priority high — Application/Job Posting Page/Function: Followup fixes for permission to create different question sets — specifically the Move Up renumbering issue re-confirmed by Aditya Dubey on Sep 9, 2026)

When clicking the Move Up arrow on a question, the question correctly moved to its new position in the list, but its displayed number stayed attached to the old position instead of recalculating (e.g. resulting in 2,1,3 instead of 1,2,3). The same incorrect numbering was also reflected in the Preview Form.

Root cause: numbers like "3.)" were literal characters manually typed into the question label text (`questionText`) — there was no live, position-based numbering anywhere in the app. Moving a question's array position obviously doesn't rewrite arbitrary text a user typed, so old numbers stayed stuck to the wrong questions after a move.

## Related PRs (if any):
Related to previous PRs #3989, #1685, PR #5103 (reviewer Deepigha's feedback on this issue).

## Main changes explained:
- Update `jobFormQuestionUtils.js` — added `stripLeadingQuestionNumber()` (strips a leading manually-typed number prefix like "13.)", "2.", "7)", "4-") and `numberedQuestionLabel()` (strips any old number and prepends the correct live number based on current array position).
- Update `JobFormbuilder.jsx` — the question list now renders `numberedQuestionLabel(field.questionText, index)` instead of raw `field.questionText`, so the number always matches current position. Also updated `editField()` to strip the number before populating the edit modal, so editing a question doesn't show or re-save a stale embedded number.
- Update `FormPreviewModal.jsx` — all 8 field-type render cases (textbox, textarea, date, file, checkbox, radio, dropdown, default) updated to use `numberedQuestionLabel()`, so the Preview Form's numbering stays in sync with the builder.
- Create `jobFormQuestionUtils.numbering.test.js` — 10 unit tests covering prefix stripping (various formats), live number generation, and a simulated reorder/swap that directly models the reported bug.

## How to test:
1. Check out this branch: `gayatri/jobformbuilder-move-up-renumbering`
2. `npm install` (or `yarn`), run frontend + backend dev servers as usual
3. Log in as a user with job-form management permissions, go to `/jobformbuilder`
4. Add at least 3 distinct questions via the Field Label input
5. Click the ↑ (Move Up) arrow on the last question — confirm it moves to the top **and** its number updates to "1.)", with the other questions renumbering accordingly
6. Repeat with ↓ (Move Down) — confirm the same correct renumbering
7. Click **Preview Form** — confirm the numbers shown there exactly match the builder's current order
8. Run unit tests: `npx vitest src/components/Collaboration/__tests__/jobFormQuestionUtils.numbering.test.js` → 10/10 should pass

## Note:
No backend changes needed — this is purely a display-layer fix. The underlying stored `questionText` (including any old baked-in numbers from before this fix) is left untouched by this change; going forward, editing a question strips old numbers so data cleans up naturally over time. No destructive rewrite of existing production data was performed.

### DIFF
--- a/src/components/Collaboration/FormPreviewModal.jsx
+++ b/src/components/Collaboration/FormPreviewModal.jsx
@@ -5,6 +5,7 @@ import {
   resolveInputType,
   STANDARD_APPLICANT_FIELDS,
   isFieldRequired,
+  numberedQuestionLabel,
 } from './jobFormQuestionUtils';
 
 /**
@@ -40,7 +41,7 @@ function FormPreviewModal({ isOpen, onClose, formFields, jobTitle, darkMode }) {
         return (
           <div key={fieldKey} className={styles.previewField}>
             <label className={styles.previewLabel}>
-              {questionText}
+              {numberedQuestionLabel(questionText, index)}
               {requiredMark}
             </label>
             <input
@@ -58,7 +59,7 @@ function FormPreviewModal({ isOpen, onClose, formFields, jobTitle, darkMode }) {
         return (
           <div key={fieldKey} className={styles.previewField}>
             <label className={styles.previewLabel}>
-              {questionText}
+              {numberedQuestionLabel(questionText, index)}
               {requiredMark}
             </label>
             <textarea placeholder="Enter text here" disabled className={styles.previewTextarea} />
@@ -69,7 +70,7 @@ function FormPreviewModal({ isOpen, onClose, formFields, jobTitle, darkMode }) {
         return (
           <div key={fieldKey} className={styles.previewField}>
             <label className={styles.previewLabel}>
-              {questionText}
+              {numberedQuestionLabel(questionText, index)}
               {requiredMark}
             </label>
             <input type="date" disabled className={styles.previewInput} />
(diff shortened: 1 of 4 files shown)


iAbhi001 (changes requested):
Thanks for working on this, @sawantgayatri19! The Move Up/Down renumbering and preview sync work great overall.

However, during testing with manually entered prefixes, I noticed an edge case with the prefix stripper `stripLeadingQuestionNumber()`.

When entering a label prefixed with `1.)` (number + period + parenthesis), the stripper only partially removes the prefix, leaving behind a trailing `.)`. This causes the dynamic label to display double punctuation like:

> **`4.) .)` What is your location? *** *(see attached screenshot)*

#### Suggested Fix:

In `jobFormQuestionUtils.js`, update the regex inside `stripLeadingQuestionNumber` so it consumes both `.` and `)` together when they follow a number (e.g., `^\s*\d+(\.\)|\.|\)|-)?\s*` or similar).

Once that regex is tweaked and tested with formats like `1.)`, this should be good to approve!

---

Image

DeepighaJ (changes requested):
- As mentioned in the previous review .) is added to the question field as partially removes the prefix.
- Verified the up /down arrows works perfectly fine while moving the questions in place respectively.

Image

Image

Image

Image

Image

Image

one-community:
Replaced by this PR: https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/5541

PR #2928, opened 2024-12-07
Jatin create different question sets

JatinAgrawal94:
# Description
Please include the exact bug/functionality description and a summary of the changes/ related issues. Please also include any other relevant motivation and context:

## Related PRS (if any):
This frontend PR is related to the [#1167](https://github.com/OneCommunityGlobal/HGNRest/pull/1167) backend PR.

## Main changes explained:
- src/components/JobFormBuilder.jsx file A for creating new job forms.

## How to test:
1. check into current branch
2. do `npm install` and `npm run start:local` to run this PR locally
3. Clear site data/cache
4. log as owner user
5. go to route `/jobformbuilder'
6. Verify Functionality as described in the video.

## Screenshots or videos of changes:
https://github.com/user-attachments/assets/7f9f9f94-8a4e-477d-aeb2-f13b80194d7e

## Note:
This PR is created for the job Application Page. This page needs to be linked to a page which is yet to made by another volunteer, which is why in the video all the API's are being tested that way.

PR #1167, opened 2023-08-23
Induja Add more specific link validation to Google Doc and Media file

Indujaks:
# Description
Add more specific link validation to Media File and GDoc links
This bug was noticed from my after reviewing pull request 980 on the front end. 
Custom Regex validator function specifically for the “Google Doc” & “Media Folder(dropbox)” needs to be created and implemented in the google doc and media folder input field. Currently, the “isValidUrl” is validating that, but only verifies that is starts with https:// etc. We need one that validates more specifically since the fields themselves are specific. This validator should be accessible in a central location such as the “utils” folder as it needs to be implemented both on the userprofile>links>edit & the otherlinks>userManagement>createNewUser modal. Images have been provided below of the target locations and the only validator function currently:

Other links > usermanagement > create new user modal
Screenshot 2023-08-23 at 12 35 25 PM

User profile > links > edit modal
Screenshot 2023-08-23 at 12 35 35 PM

Current isValidUrl function, but not specific enough for specific use cases
Screenshot 2023-08-23 at 12 35 40 PM

## Main changes explained:
1. Add specific validation regex pattern for google docs and drop box in new file under utils folder
2. Added specific validation check in editmodal instead of using a generic prototype.
3.  Included an extra step of validation while creating a new user, added a form error incase of invalid links.
…

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. log as admin user
4. Navigate to: Dashboard → Other Links → User Management → Create New User.
5. Fill in user details as per instructions.
6. Verify Admin Document and Media Folder accept valid links.
7.  Access the user profile page for the newly created user.
8. Edit User's Links:
              Navigate to the user's profile page.
9. Verify Google Doc and Media Folder Links:
              Test if only valid links are accepted for Google Doc and Media Folder fields.

## Screenshots or videos of changes:

Before: 
Earlier while creating a user profile/Editing a profile, it was accepting any valid link irrespective of it being a google doc or drop box.

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/11251850/23998d87-c1b4-45a0-a9c9-230c9e00a0d3

After: 
While creating a user profile/editing a profile, it accepts only specific valid links for google doc and drop box

https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/assets/11251850/df6f7525-bb37-415f-a7de-61fcfd271cf4

PR #5537, opened 2026-09-12
Aditya - Job Form Builder: remove field-level Clone and restore template tooltips

AdiDubbs:
# Description

Field-level **Clone** buttons could reappear after loading a template, even though question creation should remain under **Add Field**. The template-level Clone, Append, and Delete actions also lacked the explanatory tooltips required by the original task.

## Original task

On `/jobformbuilder`, remove field-level cloning, keep full-template cloning, and provide accessible guidance for Clone with Template, Append Template, and Delete Template.

- [Open master Google Doc Task 6 — clone-field regression](https://docs.google.com/document/d/1zifX1otZPl-VJLQk0sZqZbQFginRytlQ2vKHeQAh_Lo/edit?tab=t.0#bookmark=id.dov2g8hdv5ej)
- [Open master Google Doc Task 7 — template-action tooltips](https://docs.google.com/document/d/1zifX1otZPl-VJLQk0sZqZbQFginRytlQ2vKHeQAh_Lo/edit?tab=t.0#bookmark=id.u43sj556oy45)

### Master Google Doc task entries

Master Google Doc - Task 6 clone-field regression

Master Google Doc - Task 7 template-action tooltips

## Related PRS (if any):

Related historical work: PR #4475 and PR #4399. The master document links both to the original question-set work in PR #2928 and backend PR #1167. No backend changes.

## Main changes explained:

1. Removed the field-cloning callback and the per-question Clone button.
2. Kept **Add Field** for new questions and **Clone with Template** for full question sets.
3. Restored concise tooltips for Clone, Append, and Delete.
4. Made each tooltip visible on both hover and keyboard focus.
5. Added scoped tooltip styling so the guidance does not affect buttons elsewhere.

## How to test:

1. Check out this branch, install dependencies, and run `npm run start:local`.
2. Log in as an Owner and open `/jobformbuilder`.
3. Confirm individual questions have Edit, Move Up, Move Down, and Delete controls, but no Clone control.
4. Select and clone a template; confirm field-level Clone controls do not return.
5. Hover and keyboard-focus Clone with Template, Append Template, and Delete Template; confirm each tooltip appears with the correct guidance.
6. Repeat in dark mode.
7. Run the focused Collaboration tests and `npm run build`.

_Evidence sources: local QA mock using Owner-level fixture data, plus the local application running against the development backend with the Dev Admin account._

### Template clone tooltip

Template Clone Tooltip

### Field-level clone button removed

Job Form Builder - Individual Question Controls Without Clone

## Note:

The destructive Delete action remains unchanged; this PR only restores its explanatory guidance.

## Latest update — commit d27cbc1

Updated all four template-action tooltips so they remain hoverable while the pointer moves from the button onto the tooltip:

- Clone with Template
- Clear Template
- Append Template
- Delete Template

This addresses the reported hover behavior with `autohide={false}` while preserving hover and keyboard-focus triggers. All current PR checks pass. The remaining confirmation is reviewer browser retesting of the pointer transition.

### DIFF
--- a/src/components/Collaboration/JobFormbuilder.jsx
+++ b/src/components/Collaboration/JobFormbuilder.jsx
@@ -20,7 +20,6 @@ import {
   buildJobFormRequestor,
   isFieldRequired,
   normalizeQuestionForApi,
-  prepareQuestionClone,
   normalizeLoadedQuestions,
 } from './jobFormQuestionUtils';
 
@@ -143,33 +142,6 @@ function JobFormBuilder() {
   };
 
   // CRUD Functions with Dynamic Form ID
-  const cloneField = async (field, index) => {
-    const clonedField = prepareQuestionClone(field);
-    const previousFields = formFields;
-
-    const newFields = [
-      ...formFields.slice(0, index + 1),
-      clonedField,
-      ...formFields.slice(index + 1),
-    ];
-    setFormFields(newFields);
-
-    if (currentFormId) {
-      await syncFieldAction(
-        'clone question',
-        async () => {
-          await axios.post(ENDPOINTS.ADD_QUESTION(currentFormId), {
-            question: clonedField,
-            position: index + 1,
-            requestor: getRequestor(),
-          });
-          markAsSaved(newFields);
-        },
-        () => setFormFields(previousFields),
-      );
-    }
-  };
-
   const moveField = async (index, direction) => {
     const newIndex = direction === 'up' ? index - 1 : index + 1;
 
@@ -473,7 +445,6 @@ function JobFormBuilder() {
                       field={field}
                       index={index}
                       totalFields={formFields.length}
-                      onClone={cloneField}
                       onMove={moveField}
                       onDelete={deleteField}
                       onEdit={editField}
--- a/src/components/Collaboration/QuestionFieldActions.jsx
+++ b/src/components/Collaboration/QuestionFieldActions.jsx
@@ -5,7 +5,6 @@ function QuestionFieldActions({
   field,
   index,
   totalFields,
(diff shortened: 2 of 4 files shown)


shubhamjakhete (changes requested):
Requesting changes: the field-level Clone removal and tooltip restoration match the PR requirements, but the tooltip hover behavior has an accessibility issue.

`.tooltip` uses `pointer-events: none` and is only kept visible by `button:hover` / `button:focus`. As a result, moving the pointer from the button onto the tooltip causes it to disappear.

Please allow the tooltip itself to remain hoverable/visible while the pointer is over it so the restored guidance follows accessible tooltip behavior.

AdiDubbs:
@shubhamjakhete I addressed your tooltip-hover feedback in a5e1e0f. The template action tooltips now use the accessible pattern, so they remain available when moving from the trigger to the tooltip rather than relying on the prior pointer-events:none / trigger-only hover behavior. The branch was then updated with development in 4ea367d. Could you please re-review the current head?

AdiDubbs:
@shubhamjakhete I made a follow-up fix so the Clone, Clear, Append, and Delete template tooltips remain visible while the pointer moves from the button onto the tooltip. This addresses the hover behavior you flagged. Could you please re-review the updated PR?

Isha2303165 (approved):
Tested the Job Form Builder changes on the current branch. Field-level Clone actions are removed as expected, template cloning works correctly, and cloned templates do not restore individual Clone buttons. Clone, Clear, Append, and Delete template actions worked as expected, including tooltip behavior and dark mode. git diff --check is clean. Approved.

Image

